PR #6, opened 2022-05-26
Fix JWS versus JWT confusions and introduction wording

### LINKED ISSUE #5: Cite JWT as RFC 7519 and JWS as RFC 7515
The spec currently does not contain a reference for RFC 7519 (JWT), which it clearly needs to.  It also uses RFC 7515 (JWS) as the citation for JWT in several place, which is incorrect.

Please add a RFC 7519 (JWT) reference and cite it the first time that the term JSON Web Token (JWT) is used in the body of the spec (but not in the abstract).  And look at the current uses of RFC 7515 and correct those that should be RFC 7519.

selfissued:
Fixes #5

### DIFF
--- a/main.md
+++ b/main.md
@@ -43,11 +43,14 @@ documents that support selective disclosure of claim values.
 
 # Introduction {#Introduction}
 
-The JSON-based content of signed JSON Web Token documents as defined in
-[@!RFC7515] is secured against modification using digital signatures. A consumer
-of a JWT document that has checked the document's signature can safely assume
+The JSON-based claims in a signed JSON Web Token (JWT) [@!RFC7519] document
+are secured against modification using JSON Web Signature (JWS) [@!RFC7515] digital signatures.
+A consumer
+of a signed JWT document that has checked the document's signature can safely assume
 that the contents of the document have not been modified.  However, anyone
-receiving a JWT can read all contents of the document. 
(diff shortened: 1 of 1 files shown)


main.md
```diff
@@ -88,7 +91,7 @@ BASE64URL denotes the URL-safe Base64 encoding without padding as defined in
 
 # Terminology
 
- * A **SD-JWT** is a signed JWT [@!RFC7515], i.e., a JWS, that contains claims
+ * A **SD-JWT** is a signed JWT [@!RFC7519], i.e., a JWS [@!RFC7515], that contains claims
```
Sakurann, line 94: SD-JWT itself does not have to be signed. That’s why we changed SD-JWS to SD-JWT, right?

Sakurann (approved):

main.md
Sakurann, line 94:
```suggestion
 * A **SD-JWT** is a JWT [@!RFC7519], which can be signed as a JWS [@!RFC7515], that contains claims
```

alenhorvat: As mentioned in another PR:
Should we add a generic SD-JSON? This way the approach can be also used for JSON-LD or any other JSON object (signed or unsigned).

alenhorvat (approved):

PR #15, opened 2022-05-31
not concealing claim names

Sakurann:
reflecting conversation in Issuer #3 why we are not hiding claim names


danielfett:
I propose to move this to the privacy considerations.

Sakurann:
makes sense.

PR #19, opened 2022-06-01
Some minor edits.

christianpaquin:
Some minor edits:
- consistency of hyphenation in base64url-encode(d)
- use transaction instead of use case in one place because selective might be needed or not within one use case, depending on the interaction with RP
- explicitly called out period character to avoid misinterpreting as missing word
- some typos

### DIFF
--- a/main.md
+++ b/main.md
@@ -60,7 +60,7 @@ In such a use case, there has been no privacy-related concerns with existing JOS
 because when a signed JWT is one-time use, it contains only JWT claims that the user has consented
 in real time to release to the verifier. However, when a signed JWT is intended to be multi-use, 
 the ability to selectively disclose a subset of the claims depending on the verifier becomes crucial
-to ensure minimum disclosure and prevent verifier from obtaining claims irrelevant for the use case at hand. 
+to ensure minimum disclosure and prevent verifier from obtaining claims irrelevant for the transaction at hand.
 
 One example of such a multi-use JWT is a verifiable credential, or a
 tamper-evident credential with a cryptographically verifiable authorship that
@@ -73,7 +73,7 @@ well.
 
 Note: so far agreed to define holder binding (user's public key contained inside an SD-JWT) as an option.
 It is not mandatory since holder binding is use case specific and orthogonal to the general mechanism of 
-selective disclosure we are trying to define here.
+selective disclosure defined here.
 
 
 ## Conventions and Terminology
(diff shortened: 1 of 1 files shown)


Sakurann (approved):

main.md
```diff
@@ -595,7 +595,7 @@ Verifiers MUST go through (at least) the following steps before
 trusting/using any of the contents of an SD-JWT:
 
  1. Determine if holder binding is to be checked for the SD-JWT. Refer to (#holder_binding_security) for details.
- 2. Check that the presentation consists of six `.`-separated elements; if holder binding is not required, the last element can be empty.
+ 2. Check that the presentation consists of six period-separated (`.`) elements; if holder binding is not required, the last element can be empty.
```
Sakurann, line 598:
```suggestion
 2. Check that the presentation consists of four period-separated (`.`) elements; if holder binding is not required, the last element can be empty.
```

Sakurann:
changed six . separated to four per Christian's separate comment

danielfett:
> changed six . separated to four per Christian's separate comment

What did I miss? The presentation should be six dot separated elements. Three for the SD-JWT, three for the SD-JWT-R.

christianpaquin:
> What did I miss? The presentation should be six dot separated elements. Three for the SD-JWT, three for the SD-JWT-R.

Why would you need to merge both? I don't understand the proposed flow between issuer, holder, and verifier; perhaps a data flow section would help. I would think the holder gets a 3+1 part JWS with SVC (which can be `.`-concatenated). Then all you need to present to a verifier is the same data, along with some user proof key signature if the JWT is bound to a user key (for which a property is added in the SD-JWT-R, if I understood correctly; more details would be helpful here).

main.md
```diff
@@ -595,7 +595,7 @@ Verifiers MUST go through (at least) the following steps before
 trusting/using any of the contents of an SD-JWT:
 
  1. Determine if holder binding is to be checked for the SD-JWT. Refer to (#holder_binding_security) for details.
- 2. Check that the presentation consists of six `.`-separated elements; if holder binding is not required, the last element can be empty.
+ 2. Check that the presentation consists of four period-separated (`.`) elements; if holder binding is not required, the last element can be empty.
```
Sakurann, line 598:
```suggestion
 2. Check that the presentation consists of six period-separated (`.`) elements; if holder binding is not required, the last element can be empty.
```

Sakurann:
creating a separate issue on six vs four period separated elements to merge this editorial PR

PR #22, opened 2022-06-05
clarify SD-JWT-R in the intro

Sakurann:
Based on the feedback from @b---c

> https://www.ietf.org/archive/id/draft-fett-oauth-selective-disclosure-jwt-00.html#section-1-2  -> "This document also defines a format for so-called SD-JWT Releases (SD-JWT-R)."  - that sentence feels abrupt and doesn't really tell the reader anything. Consider just a bit more about what a SD-JWT-R is/does. Borrowing from https://www.ietf.org/archive/id/draft-fett-oauth-selective-disclosure-jwt-00.html#name-sd-jwt-release-sd-jwt-r maybe "This document also defines a format for so-called SD-JWT Releases (SD-JWT-R), which conveys a subset of the claim values of an SD-JWT in a verifiable way" or something along those lines.  Honestly, after reading the doc a few times, I'm unsure whether a SD-JWT-R is the encoded SVC or SD-JWT and SVC Combined Format. Or both. Or something else. And is it for releasing to the holder or the verifier or both? I think there's room for the various terms to be tightened up and clarified. And also in how they are used.

### DIFF
--- a/main.md
+++ b/main.md
@@ -53,12 +53,12 @@ can also read all of the claims.
 This document describes a format for JWTs that support selective
 disclosure (SD-JWT), enabling sharing only a subset of the claims included in 
 the original JWT instead of releasing all the claims to every verifier. 
-During issuance, SD-JWT is sent from the issuer to the holder alongside SD-JWT Salt/Value Container (SVC),
-a JSON object that contains mapping between raw claim values contained in the SD-JWT 
+During issuance, an SD-JWT is sent from the issuer to the holder alongside an SD-JWT Salt/Value Container (SVC),
+a JSON object that contains the mapping between raw claim values contained in the SD-JWT 
 and the salts for each claim value. 
 
 This document also defines a format for SD-JWT Releases (SD-JWT-R),
-which convey a subset of the claim values of an SD-JWT that the holder 
+which conveys a subset of the claim values of an SD-JWT that the holder 
 is selectively releasing to the verifier. During presentation, SD-JWT-R and SD-JWT are both sent
 to the verifier from the holder. To verify claim values received in SD-JWT-R, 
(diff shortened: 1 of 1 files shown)


main.md
```diff
@@ -53,7 +53,15 @@ can also read all of the claims.
 This document describes a format for JWTs that support selective
 disclosure (SD-JWT), enabling sharing only a subset of the claims included in 
 the original JWT instead of releasing all the claims to every verifier. 
-This document also defines a format for so-called SD-JWT Releases (SD-JWT-R).
+During issuance, SD-JWT is sent from the issuer to the holder alongside SD-JWT Salt/Value Container (SVC),
+a JSON object that contains mapping between raw claim values contained in the SD-JWT 
+and the salts for each claim value. 
+
+This document also defines a format for SD-JWT Releases (SD-JWT-R),
+which convey a subset of the claim values of an SD-JWT that the holder
```
selfissued, line 61: convey -> conveys

main.md
```diff
@@ -53,7 +53,15 @@ can also read all of the claims.
 This document describes a format for JWTs that support selective
 disclosure (SD-JWT), enabling sharing only a subset of the claims included in 
 the original JWT instead of releasing all the claims to every verifier. 
-This document also defines a format for so-called SD-JWT Releases (SD-JWT-R).
+During issuance, SD-JWT is sent from the issuer to the holder alongside SD-JWT Salt/Value Container (SVC),
```
selfissued, line 56: SD-JWT -> an SD-JWT (twice)

main.md
```diff
@@ -53,7 +53,15 @@ can also read all of the claims.
...
+a JSON object that contains mapping between raw claim values contained in the SD-JWT
```
selfissued, line 57: mapping -> the mapping

selfissued (changes requested):
This review identifies three minor syntax issues.  Once they are addressed, I'll approve.

Issue #13, opened 2022-05-30
information about the hash function


Sakurann:
information about the hash function should probably be included inside SD-JWT.

Based on @AlenHorvat 's comment to the PR https://github.com/oauthstuff/draft-selective-disclosure-jwt/pull/1#discussion_r877151757

christianpaquin:
Yes, needed indeed. This and more details about hash formatting (how to transform inputs into byte arrays to be passed to the hash function).

selfissued:
The alternative is to use the same hash function as was used for the signature.  This is done in OpenID Connect Core at https://openid.net/specs/openid-connect-core-1_0.html#ImplicitTokenValidation in the `at_hash` definition (as well as in the `c_hash` definition.  The language is:
> Hash the octets of the ASCII representation of the `access_token` with the hash algorithm specified in [JWA](https://openid.net/specs/openid-connect-core-1_0.html#JWA) [JWA] for the `alg` Header Parameter of the ID Token's JOSE Header. For instance, if the `alg` is `RS256`, the hash algorithm used is SHA-256.

PieterKas:
As discussed in the call, if we allow different hashes to be used, we may want to make a recommendation that for simplicity an implementor may want to use the same hash algorithm to minimise implementation errors.

PR #39, opened 2022-06-12
add hash_alg claim to address issue #13

Sakurann:
adding `hash_alg` SD-JWT claim 

Note: need to update the examples to include it once we agree this claim is needed (cc @danielfett )


danielfett:
We need to say in the checking algorithm that verifiers MUST check the alg and MUST only accept hash_algs they understand and deem secure.

selfissued (approved):
Assuming we're not going to use the trick that OpenID Connect uses to determine the hash algorithm implicitly, then I approve of this approach.

Issue #20, opened 2022-06-01
Example 1 JWS contains 4 parts; should be 3 at this point


christianpaquin:
The [flat SD example 1](https://github.com/oauthstuff/draft-selective-disclosure-jwt/blob/master/main.md#example-1---flat-sd-jwt) contains 4 period-separated parts. I assume the first 3 are the expected JWS ones, and the fourth one is the SVC. This is however only introduced in the [following section](https://github.com/oauthstuff/draft-selective-disclosure-jwt/blob/master/main.md#sd-jwt-and-svc-combined-format), so I think the last part should be deleted from the example.

Sakurann:
PR #46

PR #40, opened 2022-06-12
edit example 1 (Issue #20)

Sakurann:
removes SVC from the example 1 because it is introduced in the following section (Issue #20)


danielfett:
This should not be merged. The examples are generated from the python code, and here the problem was that this particular example was not properly updated by the python code. I created a fix in PR #46.

Issue #21, opened 2022-06-01
Clarify encoding of SD-JWT-R


christianpaquin:
In the [SD-JWT-R section](https://github.com/oauthstuff/draft-selective-disclosure-jwt/blob/master/main.md#sd-jwt-r-format), it is said: "the result (SD-JWT-R) is encoded as described in RFC7519"; more guidance should be provided (section number in RFC7519, serialization description, ...).

Issue #12, opened 2022-05-30
should we allow signing formats other than JWS, ie LD-Proofs, JAdES and group signatures?


Sakurann:
Do we want to allow signing formats other than JWS, ie LD-Proofs, JAdES and group signatures?

I am not an expert on JAdES, but from my understanding, I think it is close enough to JWS, it should be included. 
I do not see the benefit of including LD-Proofs and group signatures, since the whole point of this work was to enable selective disclosure without using those signature formats...

Based on the @alenhorvat 's comment to my PR: https://github.com/oauthstuff/draft-selective-disclosure-jwt/pull/1#discussion_r877144122

danielfett:
Also, we need to keep things simple. JAdES can make sense.

alenhorvat:
You're right. I missed one element:
- JWS, JAdES sign JSON as is
- LD-proofs, group signatures usually require additional transformation (and we're avoiding this).
Keeping JWS is sufficient. (JAdES results either with compact serialised or JSON serialised JWS, so we're covered).

Let's keep only JWS and we can close this issue.

selfissued:
I agree that we should support only JWS signatures as JWTs do, since our spec is explicitly a JWT profile.

I think that we should close this issue with no action on that basis.

Sakurann:
To clarify, do we agree that since signing as JWS allows for both JWS compact serialization and JWS JSON serialization, this specification supports JaDES?

selfissued:
In the name of keeping simple things simple, I would advocate supporting exactly one signing format - the JWS Compact Serialization - making the same choice as JWTs.  That will result in the greatest interoperability.

If separately, we want to create a derived spec or specs saying how to use SD-JWTs, but with other signature types, we could do that.  But that should occur in separate specification(s) - not in the SD-JWT spec itself.

Also, note that many JWS implementations only support the Compact Serialization, because that's all you need for JWTs.  The JSON Serialization is more complicated in several aspects and is much more rarely implemented.  We should not take a dependence upon it.

Sakurann:
Agreed to support only JWS Compact Serialization in SD-JWT and define a separate profile for JWS JSON serialization if needed.

@alenhorvat agrees.

PR #41, opened 2022-06-12
clarify sd-jwt-r format

Sakurann:
Clarifies text defining SD-JWT-R based on issue #21.

### DIFF
--- a/main.md
+++ b/main.md
@@ -194,7 +194,7 @@ This section defines a data format for SD-JWTs (containing hashes of the salted 
 and for SD-JWT Salt/Value Containers (containing the mapping of the plain-text claim values 
 and the salt values).
 
-## SD-JWT Format
+## Format of an SD-JWT
 
 An SD-JWT is a JWT that is optionally signed using the issuer's private key.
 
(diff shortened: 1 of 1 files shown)


main.md
```diff
@@ -484,9 +472,21 @@ KfQ
...
 
-## SD-JWT-R Format
+## Format of an SD-JWT Release
 
-The following shows the contents of an SD-JWT-R for Example 1:
+SD-JWT-R contains claim values and the salts of the claims that the holder 
+has consented to release to the Verifier. This enables the Verifier to verify 
+the claims received from the holder by computing the hash sof the claims
+values and the salts revealed in the SD-JWT-R using the hashing algorithm 
+specified in SD-JWT and comparing them to the hash valued included in SD-JWT.
```
danielfett, line 481:
```suggestion
An SD-JWT-R contains claim values and salts for the claims that the holder 
has consented to release to the Verifier. This enables the Verifier to verify 
the claims received from the holder by computing the hashes of the claim
values and the salts revealed in the SD-JWT-R using the hashing algorithm 
specified in the SD-JWT and comparing them to the hash values included in the SD-JWT.
```

main.md
```diff
@@ -499,12 +499,10 @@ The following shows the contents of an SD-JWT-R for Example 1:
...
 }
-```
-
-For each claim, an array of the salt and the claim value is contained in the
-`_sd` object. 
+``` 
 
-Again, the SD-JWT-R follows the same structure as the `_sd` in the SD-JWT. For Example 2, a SD-JWT-R limiting `address` to `region` and `country` only could look as follows:
+The following is a non-normative example of an SD-JWT-R for SD-JWT in Example 2
+that discloses only `region` and `country` of an `address` property:
```
danielfett, line 505:
```suggestion
The following is a non-normative example of an SD-JWT-R for the SD-JWT in Example 2
that discloses only `region` and `country` of the `address` property:
```

main.md
```diff
@@ -523,14 +521,16 @@ Again, the SD-JWT-R follows the same structure as the `_sd` in the SD-JWT. For E
 }
 ```
 
-The SD-JWT-R MAY contain further claims, for example, to ensure a binding
-to a concrete transaction (in the example the `nonce` and `aud` claims).
+When the holder sends SD-JWT-R to the Verifier, it MUST be a base64url-encoded 
+JWS as described in Section 5.1 of [@!RFC7515]. JWS Compact Serialization and 
+JWS JSON Serialization can be used.
```
danielfett, line 526:
```suggestion
When the holder sends SD-JWT-R to the Verifier, the SD-JWT-R MUST be a base64url-encoded 
JWS as described in Section 5.1 of [@!RFC7515]. JWS Compact Serialization and 
JWS JSON Serialization can be used.
```

main.md
```diff
@@ -523,14 +521,16 @@ Again, the SD-JWT-R follows the same structure as the `_sd` in the SD-JWT. For E
...
 
 If holder binding is desired, the SD-JWT-R is signed by the holder. If no
 holder binding is to be used, the `none` algorithm is used, i.e., the document
 is not signed.
 
-In any case, the result is encoded as described in [@!RFC7519] (here for Example 1):
+Below is a non-normative example of a representation of SD-JWT-R for SD-JWT
+given in Example 1 using JWS Compact Serialization:
```
danielfett, line 533:
```suggestion
Below is a non-normative example of a representation of the SD-JWT-R for the SD-JWT
given in Example 1 using JWS Compact Serialization:
```

danielfett (changes requested):

selfissued:
I approve of @danielfett 's suggested changes.

main.md
```diff
@@ -523,14 +521,16 @@ Again, the SD-JWT-R follows the same structure as the `_sd` in the SD-JWT. For E
 }
 ```
 
-The SD-JWT-R MAY contain further claims, for example, to ensure a binding
-to a concrete transaction (in the example the `nonce` and `aud` claims).
+When the holder sends SD-JWT-R to the Verifier, it MUST be a base64url-encoded
```
selfissued, line 524: "base64url-encode JWS" is a confusing description.  If the JWS uses the Compact Serialization, its parts are already base64url-encoded, therefore, one wouldn't re-encoded the JWS (which the current wording seems to imply).  Whereas, if it uses the JSON Serialization, then yes, one would need to base64url-encode it.  This should be clarified in the description.

main.md
```diff
@@ -523,14 +521,16 @@ Again, the SD-JWT-R follows the same structure as the `_sd` in the SD-JWT. For E
 }
 ```
 
-The SD-JWT-R MAY contain further claims, for example, to ensure a binding
-to a concrete transaction (in the example the `nonce` and `aud` claims).
+When the holder sends SD-JWT-R to the Verifier, it MUST be a base64url-encoded 
+JWS as described in Section 5.1 of [@!RFC7515]. JWS Compact Serialization and
```
selfissued, line 525: [5.1](https://datatracker.ietf.org/doc/html/rfc7515#section-5.1).  Message Signature or MAC Computation doesn't seem like the section that you meant to reference.

main.md
selfissued: Again, "base64url-encoded JWS" needs to be corrected, as described in a previous comment.

selfissued, line 526: Are we saying somewhere why are allowing the use of the JWS JSON Serialization?  On the surface of it, this will surprise people, as JWTs only permit the use of the Compact Serialization - so as to improve interoperability.  Many JWS implementations only support the Compact Serialization.

This should be discussed in its own section possibly called "JWS Serializations Supported", rather than introduced in passing, as it is in the current PR.

selfissued (changes requested):
The wording "base64url-encoded JWS" needs to be corrected, as described in the PR comments.  Also, the JWS serializations supported need to be explicitly discussed if we're going to support both serializations.

main.md
selfissued: Also, see my comment https://github.com/oauthstuff/draft-selective-disclosure-jwt/issues/12#issuecomment-1153834228 .  I really don't think we should support signing formats other than the JWS Compact Serialization in the core SD-JWT spec.  If we want to define SD-JWT-like data structures using other signing formats, that should happen in different specification(s).

Issue #25, opened 2022-06-05
same key name `_sd` in SD-JWT and SVC


Sakurann:
A comment from @b---c
> I'd think the SVC should use a different key name than "_sd". I find it confusing that the same key name is used for the selectively disclosable claims and the salt value container. More explicitly naming the container for selectively disclosable claims might be good too.

selfissued:
I agree that using different claim names should be used when their purpose is different.

peppelinux:
@Sakurann isn't it be closed thanks to https://github.com/oauthstuff/draft-selective-disclosure-jwt/pull/42?

danielfett:
Fixed in #42 indeed.

Issue #4, opened 2022-05-26
Rename sd_claims to _sd_claims


selfissued:
In OpenID Connect, we use the convention that meta claim names (claims about sets of claims) are prefixed by an underscore.  For example, see the `_claim_names` and `_claim_sources` claims defined at https://openid.net/specs/openid-connect-core-1_0.html#AggregatedDistributedClaims.  I think we should do the same.

Please rename `sd_claims` to `_sd_claims`.

danielfett:
Thanks, I renamed `sd_claims` to just `_sd`.

PR #42, opened 2022-06-12
clarify confusion around `_sd`

Sakurann:
As pointed out in the issuer #25, it is very confusing to have three separate objects that are all called `_sd`:
Renaming them as follows
`_sd` in SD-JWT -> `sd_hashes`
`_sd` in SVC -> `sd_container`
`_sd` in SD-JWT-R -> `_sd_container`

### DIFF
--- a/demo/lib.py
+++ b/demo/lib.py
@@ -11,7 +11,8 @@ from jwcrypto.jwk import JWK
 
 from .walk_by_structure import walk_by_structure
 
-SD_CLAIMS_KEY = "_sd"
+SD_DIGESTS_KEY = "sd_digests"
+SD_CLAIMS_KEY = "sd_release"
 
 # For the purpose of generating static examples for the spec, this command line
 # switch disables randomness. Using this in production is highly insecure!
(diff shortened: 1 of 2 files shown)


selfissued (changes requested):
I agree that we should differentiate between these different claims.  However, per issue #4 , we should follow the convention of using the "_" prefix in the names of all meta-claims.  Please apply the prefix, then I will approve.  (That, or convince me that some of them are not meta-claims, and I'll also approve after that discussion.)

bc-pi:
What's the reasoning for `_sd_container` and `sd_container`? They are the same thing. Just in different contexts.

danielfett:
+1 for the underscore.

'container' is not very descriptive here, what about `_sd_releases`?

bc-pi:
Aggregated and Distributed Claims are a rarely used corner of OIDC so calling that convention is a bit of an overstatement. Using an underscore as a halfhearted qualifier isn't really in the spirit of RFC7519's treatment of claim names either. 
Use an underscore if you want. But explain what it is supposed to mean. And register the claim. Don't assume it has special powers.

Sakurann:
agreed to have two containers only: sd_digests and sd_release.
Kristina to investigate wrt the usage of `_` but leaning towards not using it as a starting point.

danielfett (approved):

Issue #26, opened 2022-06-06
Clearly how a term `payload` is used.


Sakurann:
Comment from @b---c. Kristina to revise the text to make distinction between SD-JWT and SVC clearer

> The SD-JWT payload (or similar) is referred to in several examples and sometimes seems to mean the actual payload with _sd and hashes of the SD-JWT but other times seems to point to the raw/clear values of the unhidden data would be. Some different wording would be helpful to distinguish things in this regard.  https://www.ietf.org/archive/id/draft-fett-oauth-selective-disclosure-jwt-00.html#section-4.1.2 is one example of this. It seems minor but I found it to be a mental burden when reading. Like, "wait the payload of the SD-JWT has nothing but regular claims" went though my mind while reading and that makes me second guess whether I'm understanding the whole document.

PR #43, opened 2022-06-13
clarify how a term `payload` is used

Sakurann:
As pointed out in Issuer #26, a term payload has been used to characterize an object with a set of claims that the Issuer is trying to disclose using SD-JWT. Also clarifies when it is an example of an entire SD-JWT payload or just an `_sd` object.

Note: I would suggest using only Example 1 of 'flat' SD-JWT in the main text and move Example 2 to the Annex.

### DIFF
--- a/main.md
+++ b/main.md
@@ -198,11 +198,11 @@ and the salt values).
 
 An SD-JWT is a JWT that is optionally signed using the issuer's private key.
 
-### Payload
+### SD-JWT Claims
 
 The payload of an SD-JWT can consist of the following claims.
 
-#### Selectively Disclosable Claims
+#### `_sd` (Selectively Disclosable) Claim
 
 An SD-JWT MUST include hashes of the salted claim values that are included by the issuer
 under the property `_sd`. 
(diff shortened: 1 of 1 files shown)


main.md
```diff
@@ -698,7 +705,7 @@ TBD
 
 In this example, a complex object such as those used for ekyc (todo reference) is used.
 
-These claims are the payload of the SD-JWT:
+In this example, the Issuer is using a following object as a set of claims to issue to the Holder:
```
danielfett, line 708:
```suggestion
In this example, the Issuer is using the following object as a set of claims to issue to the Holder:
```

main.md
```diff
@@ -244,15 +232,23 @@ Note: need to define how holder public key is included, right now examples are u
...
+hashed claim values without any deeper structure. The `_sd` object can also be a
+'structured' object, where some claims and their respective hashes are contained
+in places deeper in the structure. it is at the issuer's discretion whether to use
+a 'flat' or 'structured' `_sd` SD-JWT object, and how to structure it such that
+it is suitable for the use case.
 
-In this example, these claims are the payload of the SD-JWT:
+Examples 1 is a non-normative example of an SD-JWT using a 'flat' `_sd` object
+and example 2 is a non-normative example of an SD-JWT using a 'structured' `_sd` object.
+The difference between the examples is how an `address` claim is disclosed.
```
danielfett, line 249:
```suggestion
The difference between the examples is how the `address` claim is disclosed.
```

danielfett (changes requested):

selfissued:
I agree with @danielfett 's suggested changes.

selfissued (approved):
I'm good to go with this once @danielfett 's suggestions have been incorporated.

selfissued (approved):
Please see the content of my previous approval.  I clicked the wrong part of the page and cleared my "reviewed" status.

bc-pi (approved):

Issue #2, opened 2022-05-19
Add example: JWT-VC as in W3C VC data model


danielfett:

tlodderstedt:
Can you please also add an example of a W3C compatible Verifiable Presentation?

PR #45, opened 2022-06-13
adding an example of a SD-JWT-R as a W3C VP in example 3

Sakurann:
per issue #2


danielfett (commented):
This seems to be based not on master, but to include some changes from other pull requests. I'll review again after the other pull requests are in master.

Muthurajj:
Where this file i could find it?

PR #61, opened 2022-06-17
chore: [demo] python code refactor

peppelinux:
Some code linting and cleanup I made reading the code and getting stuffs from it

PR #62, opened 2022-06-17
[do not merge] PoC: Use encryption instead of hashing

danielfett:
(spec not adapted yet)


peppelinux:
I was looking for this.
I'd propose a configuration paramenter to deal with both salted and encrypted payloads.

If you agree we may merge first this modest refactoring
https://github.com/oauthstuff/draft-selective-disclosure-jwt/pull/61

and also the CI to get safe against regressions, stale sources and things
https://github.com/oauthstuff/draft-selective-disclosure-jwt/pull/63

here an example of the running CI
https://github.com/peppelinux/draft-selective-disclosure-jwt/runs/6937986323?check_suite_focus=true

Sakurann:
+1 to Daniel's comment.

danielfett:
Since the spec has moved far away from what this PR was built for and problems have been addressed in other ways, I'm closing this PR.

PR #63, opened 2022-06-17
feat: GH CI

peppelinux:
Github action continuous integration tests for the youth

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -1,0 +1,45 @@
+
+name: selective-disclosure-jwt-demo
+
+on:
+  push:
+    branches: '*'
+  pull_request:
+    branches: '*'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - '3.8'
+          - '3.9'
+          - '3.10'
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install python dependencies
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/python-app.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - '3.8'
+          - '3.9'
+          - '3.10'
```
danielfett, line 21: Considering this is more a demo/reference code, not a library, I don't think we need to run this on three python versions. But I'm fine either way.

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,45 @@
...
+    # - name: Lint with flake8
+    # run: |
+    #   # stop the build if there are Python syntax errors or undefined names
+    #    flake8 ./spid_cie_oidc  --count --select=E9,F63,F7,F82 --show-source --statistics
+    #    # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+    #    flake8 ./spid_cie_oidc --count --exit-zero --statistics --max-line-length 120
+    - name: demo test
+      run: |
+        python3 -m demo.simple
+        python3 -m demo.simple_structured
```
danielfett, line 45: Adding `--replace` to the first example as that will tell us if the placeholders in main.md can be found; at least for this example, the other examples by design don't complain if the placeholders cannot be found.

Adding the third example we have.

```suggestion
        python3 -m demo.simple --replace
        python3 -m demo.simple_structured
        python3 -m demo.complex_structured
```

danielfett (changes requested):

.github/workflows/python-app.yml
peppelinux: moving on the willing to migrate this PoC to a real python package and also considering the fact that an analyst may have different python versions this is a quite standard CI

PR #78, opened 2022-06-24
Editorial suggestions

christianpaquin:
Some editorial changes:
* Use (hash) digest instead of hash when describing the result of the hash operation
* I'm suggesting to make the salt size a SHOULD vs. a MUST: 1) it doesn't affect interop, 2) some use cases might need smaller salts (with the option of using stronger (time/memory) digest derivation functions (hash)) to reduce the size of the artefact (e.g., constrained QR codes).
* `_sd` --> `sd_digests`
* Trimmed down duplicated info
* More security consideration details for hash function
* Clarified unlinkability statement in privacy consideration
* Some typos

### DIFF
--- a/main.md
+++ b/main.md
@@ -222,7 +222,7 @@ If holder binding is desired, `SD-JWT` must contain information about key materi
 SD-JWT-DOC = (METADATA, HOLDER-PUBLIC-KEY, SD-CLAIMS)
 ```
 
-Note: How the public key is included in SD-JWT is out of scope of this document. It can be passed by value or by reference. Examples in this document use `sub_jwt` Claim to include raw public key by value in SD-JWT.
+Note: How the public key is included in SD-JWT is out of scope of this document. It can be passed by value or by reference. Examples in this document use `sub_jwk` Claim to include raw public key by value in SD-JWT.
 
 With holder binding, the `SD-JWT-RELEASE` is signed by the holder using its private key. It therefore looks as follows:
 
--- a/main.md
+++ b/main.md
@@ -686,9 +686,10 @@ and revealing the claim names does not provide any additional information.
 
 ## Unlinkability 
 
-It is also important to note that this format enables selective disclosure of claims, but
-in itself it does not achieve unlinkability of the subject of an SD-SWT.
-
+Colluding issuer/verifier or verifier/verifier pairs could learn undisclosed holder claims by linking 
+issuance/presentation or two presentation sessions on the basis of unique values encoded in the SD-JWT
+(issuer signature, salts, digests, etc.) More advanced cryptographic schemes, outside the scope of
+this specification, can be used to prevent this type of linkability.
 
 # Acknowledgements {#Acknowledgements}
       
--- a/main.md
+++ b/main.md
@@ -662,14 +662,13 @@ guess.
 ## Minimum length of the salt
 
 The length of the randomly-generated portion of the salt SHOULD be at least 128 bits.
- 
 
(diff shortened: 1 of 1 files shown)


main.md
```diff
@@ -686,9 +686,10 @@ and revealing the claim names does not provide any additional information.
 
 ## Unlinkability 
 
-It is also important to note that this format enables selective disclosure of claims, but
-in itself it does not achieve unlinkability of the subject of an SD-SWT.
-
+Colluding issuer/verifier or verifier/verifier pairs could learn undisclosed holder claims by linking
```
sakimura, line 689: Learning of undisclosed holder claims does not sound like an unlinkability issue. 
In the case of (verifier1+verifier2)-holder1-unlinkability, it is the linking of holder1's presentations (protocol executions) at verifier1 and verifier2. This is the failure of the pairwise pseudonymity property. 

If we want to talk about the uncovering of claims by colluding issuer and verifier, then we probably want to have another section on it, and IMHO, we should have one.

main.md
```diff
@@ -687,9 +686,10 @@ and revealing the claim names does not provide any additional information.
 
 ## Unlinkability 
 
-It is also important to note that this format enables selective disclosure of claims, but
-in itself it does not achieve unlinkability of the subject of an SD-SWT.
-
+Colluding issuer/verifier or verifier/verifier pairs could learn undisclosed holder claims by linking 
+issuance/presentation or two presentation sessions on the basis of unique values encoded in the SD-JWT
+(issuer signature, salts, digests, etc.) More advanced cryptographic schemes, outside the scope of
+this specification, can be used to prevent this type of linkability.
```
Sakurann, line 692: Happy to approve if the change in line to the one suggested below will be incorporated.

```suggestion
Colluding issuer/verifier or verifier/verifier pairs could link issuance/presentation or two presentation sessions
to the same user on the basis of unique values encoded in the SD-JWT
(issuer signature, salts, digests, etc.). More advanced cryptographic schemes, outside the scope of
this specification, can be used to prevent this type of linkability.
```

main.md
Sakurann: I made a suggestion wrt unlinkability text. I agree learning undisclosed holder claims is a separate issue that we should probably have a text about in security considerations, outside of this PR.

main.md
```diff
@@ -253,23 +253,24 @@ subset of the same mapping).
 
 An SD-JWT is a JWT that MUST be signed using the issuer's private key. The
 payload of an SD-JWT MUST contain the `sd_digests` and `hash_alg` claims
-described in the following, and MAY contain a holder's public key or a reference
+described in the following sections, and MAY contain a holder's public key or a reference
```
danielfett, line 256: I chose "in the following" for being less ambiguous, as "following section" can be read as "go to the next top-level section to read more", but we want to say "in the following subsections", which is an awkward way to word it. Therefore just "in the following". But since this is the second time someone suggests this edit, I'm starting to question my interpretation here....

danielfett (changes requested):

main.md
Sakurann: let's just say `in the following subsections` or `in this document` than!

main.md
christianpaquin: I mean claim "values" here, not their type. Linkability can be established on the basis of any correlatable information, which could be any unique value (serial number, signature value, a claim value).

danielfett (approved):

Sakurann (approved):

Issue #60, opened 2022-06-17
Add proper header claims in SD-JWT


peppelinux:
We have 

````
{
  "alg": "RS256"
}
````

we may have something like

````
{
  "typ": "SD-JWT"
  "alg": "RS256"
}
````

tlodderstedt:
+1

peppelinux:
I suppose also to have at least `kid` or `x5t` or `jwk` or `x5c` in the headers.

peppelinux:
and also the typ header claim "SD-JWT-R"

do we have nay prefernces on the name of these typ(s)?

"selective-disclosure-jwt+r" and things or the KISS approach? (I'm for the second one)

danielfett:
If I look at what we have chosen for DPoP, I think we should go for `sd+jwt` and `sd-r+jwt` or similar. @b---c Do you have an opinion on this?

bc-pi:
Agree with `sd+jwt`. 

Not sure about the `sd-r+jwt` because the SD JWT release is more stuff than a single JWT. And typ is describing the JWT containing it.

peppelinux:
@bc-pi

This is the SD-JWT-R taken from draft and demo code
````
eyJhbGciOiAiUlMyNTYifQ.eyJub25jZSI6ICIyR0xDNDJzS1F2ZUNmR2ZyeU5STjl3Iiw
gImF1ZCI6ICJodHRwczovL2V4YW1wbGUuY29tL3ZlcmlmaWVyIiwgIl9zZCI6IHsiZ2l2Z
W5fbmFtZSI6ICJbXCI2SWo3dE0tYTVpVlBHYm9TNXRtdlZBXCIsIFwiSm9oblwiXSIsICJ
mYW1pbHlfbmFtZSI6ICJbXCJlSThaV205UW5LUHBOUGVOZW5IZGhRXCIsIFwiRG9lXCJdI
iwgImFkZHJlc3MiOiAiW1wiUGMzM0pNMkxjaGNVX2xIZ2d2X3VmUVwiLCB7XCJzdHJlZXR
fYWRkcmVzc1wiOiBcIjEyMyBNYWluIFN0XCIsIFwibG9jYWxpdHlcIjogXCJBbnl0b3duX
CIsIFwicmVnaW9uXCI6IFwiQW55c3RhdGVcIiwgXCJjb3VudHJ5XCI6IFwiVVNcIn1dIn1
[...]

PR #83, opened 2022-06-25
feat: [Draft] added sd+jwt header for SD-JWT

peppelinux:
Definition of these new JWT headers with normative language.
Closes https://github.com/oauthstuff/draft-selective-disclosure-jwt/issues/60

### DIFF
--- a/main.md
+++ b/main.md
@@ -257,11 +257,7 @@ described in the following, and MAY contain a holder's public key or a reference
 thereto, as well as further claims such as `iss`, `iat`, etc. as defined or
 required by the application using SD-JWTs.
 
-The SD-JWTs MUST be explicitly typed, by setting the `typ` header 
-parameter to `sd+jwt`. This is done to prevent cross-JWT 
-confusion (see "RFC8725", section 3.11).
-
-
+A `typ` parameter in the JOSE header of an SD-JWT MUST be set to the value `sd+jwt`.
 ### `sd_digests` Claim (Digests of Selectively Disclosable Claims)
 
 An SD-JWT MUST include hashes of the salted claim values that are included by the issuer
--- a/main.md
+++ b/main.md
@@ -609,13 +609,13 @@ trusting/using any of the contents of an SD-JWT:
  2. Check that the presentation consists of six period-separated (`.`) elements; if holder binding is not required, the last element can be empty.
  3. Separate the SD-JWT from the SD-JWT Release.
  4. Validate the SD-JWT:
-    1. Ensure that a signing algorithm was used that was deemed secure for the application. Refer to [@RFC8725], Sections 3.1 and 3.2 for details.
-    2. Validate the signature over the SD-JWT. 
-    3. Validate the issuer of the SD-JWT and that the signing key belongs to this issuer.
-    4. Check that the SD-JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the SD-JWT.
-    5. Check that the claim `sd_digests` is present in the SD-JWT.
-    6. Check that the `hash_alg` claim is present and its value is understand
-       and the hash algorithm deemed secure.
+    1. Check that the `typ` JOSE header parameter has the value sd+jwt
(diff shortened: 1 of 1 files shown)


main.md
```diff
@@ -257,6 +257,11 @@ described in the following, and MAY contain a holder's public key or a reference
 thereto, as well as further claims such as `iss`, `iat`, etc. as defined or
 required by the application using SD-JWTs.
 
+The SD-JWTs MUST be explicitly typed, by setting the `typ` header 
+parameter to `sd+jwt`. This is done to prevent cross-JWT 
+confusion (see "RFC8725", section 3.11).
+
+
```
Sakurann, line 264:
```suggestion
A `typ` parameter in the JOSE header of an SD-JWT MUST be set to the value `sd+jwt`.
```

Also can you please update a validation section with something like this?

```
 4. Validate the SD-JWT:
    1. Check that the `typ` JOSE header parameter has the value sd+jwt
    2. Ensure that a signing algorithm was used that was deemed secure for the application. Refer to [@RFC8725], Sections 3.1 and 3.2 for details.
    3. Validate the signature over the SD-JWT. 
    4. Validate the issuer of the SD-JWT and that the signing key belongs to this issuer.
    5. Check that the SD-JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the SD-JWT.
    6. Check that the claim `sd_digests` is present in the SD-JWT.
    7. Check that the `hash_alg` claim is present and its value is understand
       and the hash algorithm deemed secure.
```

main.md
```diff
@@ -503,6 +508,10 @@ When the holder sends the SD-JWT-R to the Verifier, the SD-JWT-R MUST be a JWS
 represented as the JWS Compact Serialization as described in 
 Section 7.1 of [@!RFC7515].
 
+The SD-JWT-R MUST be explicitly typed, by setting the `typ` header 
+parameter to `sd-r+jwt`. This is done to prevent cross-JWT 
+confusion (see "RFC8725", section 3.11).
+
```
Sakurann, line 514: Note that SD-JWT-R is not always signed.

```suggestion
When SD-JWT-R is signed, a `typ` parameter in the JOSE header of a signed SD-JWT MUST be set to the value `sd-r+jwt`.
```

Sakurann (changes requested):
I would be more comfortable adding only SD+JWT typ for now, since we still need to agree on SD-JWT-R being a JWT vs an object and its validation logic.

peppelinux:
> I would be more comfortable adding only SD+JWT typ for now, since we still need to agree on SD-JWT-R being a JWT vs an object and its validation logic.

Ok.
If sd-jwt-r would be an object and not a jwt how we could bind the sub_jwk of sd-jwt, issued by a trusted party (vci) to the selfsigned release issued by the holder?

Sakurann:
SD-JWT-R is an object when there is no holder binding ie no`sub_jwk` inside SD-JWT.

bc-pi:
see https://github.com/oauthstuff/draft-selective-disclosure-jwt/issues/60#issuecomment-1167745219

b---c (changes requested):
shouldn't do this

peppelinux:
> SD-JWT-R is an object when there is no holder binding ie no`sub_jwk` inside SD-JWT.

mmmhh ... anyway it doesn't have any sense due to the possibility that the SD-JWT can be stolen. The signature and the binding of the sub jwk is a concrete proof of possession

main.md
peppelinux: Done here
https://github.com/oauthstuff/draft-selective-disclosure-jwt/pull/83/commits/a4e455d62a6baaa879739975fd50f22256927da6

thank you, it was really important

peppelinux:
SD-JWR-R typ header removed here
https://github.com/oauthstuff/draft-selective-disclosure-jwt/pull/83/commits/b89b4aca8f46cafc7ebfebe69003cf8253617e1f

peppelinux:
Not of interest anymore

Issue #80, opened 2022-06-24
[Privacy] Anonymous claims


peppelinux:
### Problem
A user doesn't want to show which attributes he got from an issuer in a VC. The reason behind this is that the user doesn't want to give to a RP any information about the attributes that this RP may ask to him in the future. If a RP has a proof of possession of some claim of a user it may ask these to the user, in the future, producing  some reason, some levers or circumstances, to persuade a need to the user to share his claims to the RP.

A RP may know that a user is also:
- graduated even without knowing the value contained in the related claim.
- married, without knowing the attributes of the spouse
- and so on ...

This means that the only presence of a claim, even if its value is opaque, can determine an information or a proof of truth.
This scenario is not relevant for a person schema (eIDAS minimum data set) but can be relevant with the issuing of some extendend attestation of the personal attributes.

### Requirement
A way to make also the claim names anonymous/opaques.

### Current behaviour
The **sd_digest** shows how many and which attributes the user may disclose

````
...
  "sd_digests": {
    "sub": "LbnhkOr5oS7KjeUrxezAu8TG0CpWz0jSixy6tffuo04",
    ...
  }
...
````
### Expected behaviour
Also the claim names can be obscured, defining an additional structure type to the current three available to date (simple, structured, complex). The result can be the SD-JWT in the example given above

````
{
  "iss": "https://example.com/issuer",
  "sub_jwk": {
    "kty": "RSA",
    "n": "pm4bOHBg-oYhAyPWzR56AWX3rUIXp11_ICDkGgS6W3ZWLts-hzwI3x65659kg4hVo9dbGoCJE3ZGF_eaetE30UhBUEgpGwrDrQiJ9zqprmcFfr3qvvkGjtth8Zgl1eM2bJcOwE7PCBHWTKWYs152R7g6Jg2OVph-a8rq-q79MhKG5QoW_mTz10QT_6H4c7PjWG1fjh8hpWNnbP_pv6d1zSwZfc5fl6yVRL0DV0V3lGHKe2Wqf_eNGjBrBLVklDTk8-stX_MWLcR-EGmXAOv0UBWitS_dXJKJu-vXJyw14nHSGuxTIK2hx1pttMft9CsvqimXKeDTU14qQL1eE7ihcw",
    "e": "AQAB"
  },
  "hash_alg": "sha-256",
  "iat": 1516239022,
  "exp": 1516247022,
  "sd_digests": [
[...]

PR #93, opened 2022-06-26
Update main.md

sakimura:
To address issue #91

### DIFF
--- a/main.md
+++ b/main.md
@@ -679,16 +679,136 @@ TBD
 
 # Privacy Considerations {#privacy_considerations}
 
-## Claim Names
+Privacy protection can be considered from the following principles per ISO/IEC 29100.  
 
-Claim names are not hashed in the SD-JWT and are used as keys in a key-value pair, where the value is the hash.
-This is because SD-JWT already reveals information about the issuer and the schema,
-and revealing the claim names does not provide any additional information.
+	1.	Consent and choice 
+	2.	Purpose legitimacy and specification 
+	3.	Collection limitation 
+	4.	Data minimization 
+	5.	Use, retention and disclosure limitation 
+	6.	Accuracy and quality 
+	7.	Openness, transparency and notice 
+	8.	Individual participation and access 
+	9.	Accountability 
+	10.	Information security 
+	11.	Privacy compliance 
 
-## Unlinkability 
+Among them, this document is concerned directly with 3. Collection limitation and indirectly/partially with 1. Consent and choice, 2. Purpose legitimacy and specification, 6. Accuracy and quality, and 10. Information security.  
 
-It is also important to note that this format enables selective disclosure of claims, but
-in itself it does not achieve unlinkability of the subject of an SD-SWT.
+Often talked unlinkability is a subset of 3. Collection limitation as it amounts to providing unintended (by the holder) information such as the fact that the holder visited a verifier to another verifier. 
+
+## Collection limitation
+
+In this document, Collection limitation is considered from two aspects. 
+
+1. Uncoverability
+2. Unlinkability
+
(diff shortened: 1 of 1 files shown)


main.md
```diff
@@ -679,16 +679,137 @@ TBD
...
+
+To mitigate this kind of issue, the group of issuers may elect to use a common set of claims and always include them in the SD-JWT. 
+
+#### Uncovering the claim value where the number of possible values is limited
+
+When the number of possible values of the claim is limited, it is trivial to uncover the actual value from the salt and the hash. 
+It suffices to iterate over the possible values with the provided hash. Therefore, it is critical to keep the secrecy of 
+the corresponding hash value for the claims of which values are not disclosed. This may be harder to do than said 
+as the hash values are sent to various parties and hashed claims values are available to all the parties that received SD-JWT-R. 
+Dynamically generating the claim set that only include the claims of which values are to be exposed is safer.
```
danielfett, line 735: Isn't this covered by the salt?

main.md
```diff
@@ -679,16 +679,137 @@ TBD
...
+Then, by just looking at the issuer, it is possible to deduce that the holder is a cancer patient. 
+This may have a fair amount of privacy impact. 
+
+To mitigate this kind of issue, a group of issuers may elect to use a common issuer identifier and use a group signature scheme instead of an individual signature. 
+
+#### Uncovering the values of non-existent claims from the list of claim names
+Even if a group of issuers uses a common issuer identifier, if the claims used are unique from one issuer to another, 
+the list of the claim names would be enough to identify the real issuer. In this case, the values of non-existent claims can be uncovered in the same manner as described in the previous subclause. 
+
+To mitigate this kind of issue, the group of issuers may elect to use a common set of claims and always include them in the SD-JWT.
```
danielfett, line 727: I think it is worth discussing this approach also independent of the problem of identifying the issuer. Always including some claims prevents verifiers from learning something about the holder just from the presence of a claim.

Example: In EU driver's licenses, there may be restrictions such as "may only drive with hearing aids" or "may only drive with modified vehicles to accommodate a certain disability". The verifier may learn just from the presence of any restrictions that the owner of the license may not be of best health, potentially using it to their disadvantage.

Always including claims and just nulling the value when not applicable can prevent this.

main.md
```diff
@@ -679,16 +679,137 @@ TBD
 
 # Privacy Considerations {#privacy_considerations}
 
-## Claim Names
+Privacy protection can be considered from the following principles per ISO/IEC 29100.
```
danielfett, line 682: Is there a good link we can provide here? I fell that having the full list makes sense only if it is not readily available somewhere else.

main.md
```diff
@@ -679,16 +679,137 @@ TBD
...
+	9.	Accountability 
+	10.	Information security 
+	11.	Privacy compliance 
 
-## Unlinkability 
+Among them, this document is concerned directly with 3. Collection limitation and indirectly/partially with 1. Consent and choice, 2. Purpose legitimacy and specification, 6. Accuracy and quality, and 10. Information security.  
 
-It is also important to note that this format enables selective disclosure of claims, but
-in itself it does not achieve unlinkability of the subject of an SD-SWT.
+Often talked unlinkability is a subset of 3. Collection limitation as it amounts to providing unintended (by the holder) information such as the fact that the holder visited a verifier to another verifier.
```
danielfett, line 698: Can we specify "often talked"?

main.md
peppelinux: The following proposal resolve by design the problem of the claim names disclosed in SD-JWT
https://github.com/oauthstuff/draft-selective-disclosure-jwt/issues/80

PR #112, opened 2022-07-07
fix: small third-person typo in introduction

peppelinux:


danielfett:
The current spelling is correct since the 'convey' refers to the plural 'SD-JWT-Releases'.

That said, the sentence is not easy to read and may need improvement.

PR #117, opened 2022-07-11
OO interface for our library

danielfett:
This commit changes the interface of the library from a set of functions to an object-oriented interface.

This will help with future improvements (blinded claim names, for example).


peppelinux (approved):
It was a great pleasure to me have read this PR, this OOP approach is the natural evolution of the code and it's the best improvement to facilitate unit testing

ad maiora @danielfett

Sakurann (approved):
Agreed. this is great. thank you, Daniel.

PR #126, opened 2022-07-14
Remove outdated HTML versions

danielfett:
It's on the IETF website after all, and viewing main.md works just fine.


Sakurann:
umm... many implementors of sd-jwt as a vc would not know ietf process and where to look for the ietf htmls. I would prefer keeping main.md as-is

peppelinux:
@Sakurann the problem is that html is very outdated and we don't have a CD to keep it updated and, anyway, it would be an online replica. Having in the README a clear pointer to the IETF official page of SD-JWT I don't feel any worry for our implementers

Sakurann:
ah ok. I am ok deleting [github.io](https://oauthstuff.github.io/draft-selective-disclosure-jwt/draft-fett-selective-disclosure-jwt-00.html) for an HTML version.
but why do we need to remove [IETF OAuth WG](https://datatracker.ietf.org/doc/draft-fett-oauth-selective-disclosure-jwt/)?

README.md
```diff
@@ -5,10 +5,9 @@ documents that support selective disclosure of claim values.
 
 Written in markdown for the [mmark processor](https://github.com/mmarkdown/mmark).
 
-This is a GitHub repository for a draft specification in [IETF OAuth WG](https://datatracker.ietf.org/doc/draft-fett-oauth-selective-disclosure-jwt/)
-For the latest version of the IETF draft, please see https://datatracker.ietf.org/doc/html/draft-fett-selective-disclosure-jwt
+This is a GitHub repository for a draft specification in the IETF OAuth WG. For the latest published version of this IETF draft, please see https://datatracker.ietf.org/doc/html/draft-fett-selective-disclosure-jwt
```
peppelinux, line 8: hem.. probably @Sakurann is right, can we reenable the [IETF OAuth WG](https://datatracker.ietf.org/doc/draft-fett-oauth-selective-disclosure-jwt/) link?

as it is in this change the url is published quite raw

Sakurann, line 8:
```suggestion
This is a GitHub repository for a draft specification in the [IETF OAuth WG](https://datatracker.ietf.org/doc/draft-fett-oauth-selective-disclosure-jwt/). For the latest published version of this IETF draft, please see https://datatracker.ietf.org/doc/html/draft-fett-selective-disclosure-jwt
```

danielfett: You're right, I added the link again!

peppelinux (approved):

Sakurann (approved):

PR #145, opened 2022-09-30
Editorial: Update Document History after WG adoption

danielfett:


main.md
```diff
@@ -747,28 +747,6 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
 
 TBD
 
-<reference anchor="OIDC" target="https://openid.net/specs/openid-connect-core-1_0.html">
```
Sakurann, line 750: why are we removing this?

danielfett: It was not used anywhere in the document.

PR #147, opened 2022-10-01
[ed] added definition of selective disclosure

Sakurann:


main.md
```diff
@@ -126,6 +126,9 @@ Holder
 Verifier 
 :  An entity that requests, checks and extracts the claims from SD-JWT-R (2.2)
 
+Selective disclosure
+: Process of a Holder disclosing to the Verifier a subset of claims contained in a claim set issued by the Issuer.
```
danielfett, line 130: Probably stylistic, but this feels more natural to me:
```suggestion
: Process of a Holder disclosing to a Verifier a subset of claims contained in a claim set issued by an Issuer.
```

PR #154, opened 2022-10-14
fix: CI

peppelinux:
This PR fixes the CI that fails after the main file was renamed


danielfett:
Thank you, but I have fixed this in https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/152 already, which we'll merge soon.

Issue #90, opened 2022-06-26
Terms and definition formatting error


sakimura:
# Comments
While some of the definitions refer to other terms with numbers, the corresponding term is not numbered. 

# Proposals
Add numbers and reformat.

danielfett:
I think we should remove the references with numbers since they are not required anyway.

The format we're using for the terms is a somewhat official one, using features provided by xml2rfc. Not sure if it is the best one, though.

sakimura:
Removing the numbers is one way to remove the inconsistency and I am fine with it. It should be one way or another and be consistent. 

Having said that, the current format is hard to read, unfortunately. 

Also, I know it is not a norm in IETF, but relying on upper-casing for defined words and keywords is a bad idea, as they often do not translate. So, we should make sure that the defined terms would not appear in the main text in a dictionary meaning regardless of the case.

danielfett:
We switched to a format that is also used in other published RFCs. The new format has a bit too much wide space, but otherwise works fine.

PR #163, opened 2022-10-24
Formatting

Sakurann:


Sakurann:
Issue #90

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -122,40 +125,40 @@ Section 2 of [@!RFC7515].
...
 
-Holder
-:  An entity that received SD-JWTs from the Issuer and has control over them.
+Holder:
+  An entity that received SD-JWTs from the Issuer and has control over them.
 
-Verifier
-:  An entity that requests, checks and extracts the claims from HS-Disclosures JWT.
+Verifier:
+  An entity that requests, checks and extracts the claims from HS-Disclosures JWT.
```
danielfett, line 161: This is not the proper formatting. The formatting we're currently using is correct in terms of markdown (see https://www.markdownguide.org/extended-syntax/#definition-lists) and it gets rendered correctly into the XML, but xml2rfc does not seem to do a good job at rendering it. We should look into that.

danielfett (changes requested):

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: I checked again and it seems that others (e.g., RFC9000) also use definition lists, but add another colon after the term to be defined.

Sakurann: > Verifier:
>:  An entity that requests, checks and extracts the claims from HS-Disclosures JWT.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -89,24 +89,27 @@ cases as well.
 
 This document also describes an optional mechanism for Holder Binding,
 or the concept of binding an SD-JWT to key material controlled by the
-Holder.
+Holder. Strenght of the Holder Binding is conditional upon the trusts
+in protection of the private key of the key matetial an SD-JWT is bound to.
```
danielfett, line 93:
```suggestion
Holder. The strength of the Holder Binding is conditional upon the trust
in the protection of the private key of the key pair an SD-JWT is bound to.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1301,7 +1305,11 @@ TBD
 
 All of the following examples are non-normative.
 
-## Example 2a - Structured SD-JWT
+## Structured SD-JWT
+
+The following examples show the variations of a structured SD-JWT.
```
danielfett, line 1310:
```suggestion
The following examples show variations of a structured SD-JWT.
```

danielfett (changes requested):

danielfett:
Also this doesn't compile due to trailing spaces.

danielfett (approved):

PR #164, opened 2022-10-25
updated a linkability section based on the UL feedback

Sakurann:
Added an option to do pairwise SD-JWTs to prevent Verifier/Verifier linkability

Original feedback.

> The discussion of linkability in section 8.2 is incomplete. It actually gives the impression that linkability is unavoidable if this RFC is implemented, since for a solution it only refers to undefined ‘advanced cryptographic schemes outside the scope of this specification’. However, in reality an issuer can completely avoid linkability on the basis of unique values encoded in the SD-JWT, simply by ensuring that each SD-JWT can be used for only one transaction. Of course, this places stringent demands on the connectivity of the holder app and the capacity of the issuer’s systems for generating and communicating new SD-JWTs to its holder apps. To lessen these burdens, the issuer could also choose other approaches, for example 
> - Equipping each holder app with a pool of (for instance) 20 different SD-JWTs. The app would then randomly pick one of these SD-JWTs for every transaction, and re-start the pool once all JWTs have been used. 
> - Allowing each SD-JWT to be used multiple times, but only for a limited time.
> 
> The SD-JWT mechanism therefore allows every issuer to find the correct balance between holder linkability and mitigation costs and efforts, based on a risk analysis for their specific situation.
Please re-write section 8.2 in the light of the above suggestions.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1174,10 +1174,13 @@ Furthermore, re-using the same value to blind a claim name may limit the privacy
 
 ## Unlinkability
 
-Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation or two presentation sessions
-to the same user on the basis of unique values encoded in the SD-JWT
-(Issuer signature, salts, digests, etc.). More advanced cryptographic schemes, outside the scope of
-this specification, can be used to prevent this type of linkability.
+Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation 
+or two presentation sessions to the same user on the basis of unique values encoded in the SD-JWT
+(Issuer signature, salts, digests, etc.). 
+
+To prevent these types of linkability, various methods, including but not limited to the following ones can be used:
+- Use advanced cryptographic schemes, outside the scope of this specification.
+- Issue a barch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
 
 # Acknowledgements {#Acknowledgements}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1174,23 +1174,29 @@ Furthermore, re-using the same value to blind a claim name may limit the privacy
...
-Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation or two presentation sessions
-to the same user on the basis of unique values encoded in the SD-JWT
-(Issuer signature, salts, digests, etc.). More advanced cryptographic schemes, outside the scope of
-this specification, can be used to prevent this type of linkability.
+Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation 
+or two presentation sessions to the same user on the basis of unique values encoded in the SD-JWT
+(Issuer signature, salts, digests, etc.). 
+
+To prevent these types of linkability, various methods, including but not limited to the following ones can be used:
+- Use advanced cryptographic schemes, outside the scope of this specification.
```
danielfett, line 1182:
```suggestion
To prevent these types of linkability, various methods, including but not limited to the following ones can be used:

- Use advanced cryptographic schemes, outside the scope of this specification.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1174,23 +1174,29 @@ Furthermore, re-using the same value to blind a claim name may limit the privacy
...
+- Issue a barch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
```
danielfett, line 1183:
```suggestion
- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
```

danielfett (changes requested):

danielfett (approved):

PR #168, opened 2022-11-06
validate how HS-disclosures JWT is validated

### LINKED ISSUE #166: validate what? -01
https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-01.html#section-6.2-3.5.1 has, "Validate the HS-Disclosures JWT: ... If Holder Binding is required, validate the signature over the SD-JWT using the same steps as for the SD-JWT plus the following steps:", which can't be right with both mentions of SD-JWT.  

Furthermore I'd strongly suggest that validation steps for the HS-Disclosures JWT be described independently.  Things like "using the same steps as for the SD-JWT" seem likely to lead to confusion and/or require extrapolation (which steps exactly? the issuer step above is kinda about the signature but not really needed in HS-Disclosures).

Sakurann:
resolves #166 by clarifying HS-disclosures JWT validation and separating it into a separate section

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -916,49 +916,84 @@ The Holder SHOULD verify the binding between SD-JWT and II-Disclosures Object by
 
 ## Verification by the Verifier when Receiving SD-JWT and Holder-Selected Disclosures JWT {#verifier-verification}
 
-Verifiers MUST follow [@RFC8725] for checking the SD-JWT and, if signed, the
+follow [@RFC8725] for checking the SD-JWT and, if signed, the
 HS-Disclosures JWT.
 
 Verifiers MUST go through (at least) the following steps before
-trusting/using any of the contents of an SD-JWT:
+trusting/using any of the contents of an SD-JWT.
 
- 1. Determine if Holder Binding is to be checked according to the Verifier's policy
+If any of the following steps fail, the input is not valid and
+processing MUST be aborted.
+
+### Step 1: Validate Combined Format for Presentation
+
+1. Determine if Holder Binding is to be checked according to the Verifier's policy
     for the use case at hand. This decision MUST NOT be based on whether
     the HS-Disclosures
     JWT is signed or not. Refer to (#holder_binding_security) for
     details.
- 2. Check that the presentation consists of six period-separated (`.`) elements; if Holder Binding is not required, the last element can be empty.
- 3. Separate the SD-JWT from the HS-Disclosures JWT.
- 4. Validate the SD-JWT:
-    1. Ensure that a signing algorithm was used that was deemed secure for the application. Refer to [@RFC8725], Sections 3.1 and 3.2 for details. `none` MUST NOT be accepted.
-    2. Validate the signature over the SD-JWT.
-    3. Validate the Issuer of the SD-JWT and that the signing key belongs to this Issuer.
-    4. Check that the SD-JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the SD-JWT.
-    5. Check that the claim `sd_digests` is present in the SD-JWT.
-    6. Check that the `sd_digest_derivation_alg` claim is present and its value is understood and the digest derivation algorithm is deemed secure.
- 5. Validate the HS-Disclosures JWT:
-    1. If Holder Binding is required, validate the signature over the HS-Disclosures JWT using the same steps as for the SD-JWT plus the following steps:
-      1. Determine that the public key for the private key that used to sign the HS-Disclosures JWT is bound to the SD-JWT, i.e., the SD-JWT either contains a reference to the public key or contains the public key itself.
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -916,49 +916,84 @@ The Holder SHOULD verify the binding between SD-JWT and II-Disclosures Object by
...
 
 Verifiers MUST go through (at least) the following steps before
-trusting/using any of the contents of an SD-JWT:
+trusting/using any of the contents of an SD-JWT.
 
- 1. Determine if Holder Binding is to be checked according to the Verifier's policy
+If any of the following steps fail, the input is not valid and
+processing MUST be aborted.
+
+### Step 1: Validate Combined Format for Presentation
```
bc-pi, line 928: Stepping back a bit, I don't think that mandating that the disclosures go in a JWT is necessary. And maybe not even really appropriate. I was hoping to discuss the general issue/idea during the meeting(s) this week. But it's leading to a lot of unnecessary confusion/complexity in these steps. So it's hard to review or comment without raising it.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -916,49 +916,84 @@ The Holder SHOULD verify the binding between SD-JWT and II-Disclosures Object by
...
+4. Check that the SD-JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the SD-JWT.
+5. Check that the claim `sd_digests` is present in the SD-JWT.
+6. Check that the `sd_digest_derivation_alg` claim is present and its value is understood and the digest derivation algorithm is deemed secure.
+
+### Step 3: Validate the HS-Disclosures JWT:
+
+When there is no Holder Binding check:
+
+1. Check that the HS-Disclosures JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the HS-Disclosures JWT.
+2. Determine that the HS-Disclosures JWT is bound to the current transaction and was created for this Verifier (replay protection). This is usually achieved by a `nonce` and `aud` field within the HS-Disclosures JWT.
```
bc-pi, line 953: why/how is this going to work when there's no holder binding?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -916,49 +916,84 @@ The Holder SHOULD verify the binding between SD-JWT and II-Disclosures Object by
...
+3. For each claim in `sd_hs_disclosures` in the HS-Disclosures JWT:
```
bc-pi, line 954: this is really part of the SD-JWT validation and needs to happen regardless of holder binding or no holder binding

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -916,49 +916,84 @@ The Holder SHOULD verify the binding between SD-JWT and II-Disclosures Object by
...
+  1. Ensure that the claim is present as well in `sd_digests` in the SD-JWT.
+     If `sd_digests` is structured, the claim MUST be present at the same
+     place within the structure.
+  2. Compute the base64url-encoded digest of the JSON literal disclosed
```
bc-pi, line 958: Digest comparison should probably be using the binary. At least not mandating comparison of the encoded values as this kinda suggests.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -916,49 +916,84 @@ The Holder SHOULD verify the binding between SD-JWT and II-Disclosures Object by
...
        validated and used according to the requirements of the application. It
        MUST be ensured that all claims required for the application have been
        disclosed.
 
-If any step fails, the input is not valid and processing MUST be aborted.
+When there is Holder Binding check:
+
+1. Ensure that a signing algorithm used was deemed secure for the application. Refer to [@RFC8725], Sections 3.1 and 3.2 for details. `none` MUST NOT be accepted.
+2. Validate the signature over the HS-Disclosures JWT.
+3. Validate the Issuer of the HS-Disclosures JWT and that the signing key belongs to this Issuer.
```
bc-pi, line 976: how does this work? the issuer would be the holder in this context. there likely isn't an iss

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -916,49 +916,84 @@ The Holder SHOULD verify the binding between SD-JWT and II-Disclosures Object by
 
 ## Verification by the Verifier when Receiving SD-JWT and Holder-Selected Disclosures JWT {#verifier-verification}
 
-Verifiers MUST follow [@RFC8725] for checking the SD-JWT and, if signed, the
+follow [@RFC8725] for checking the SD-JWT and, if signed, the
```
danielfett, line 919: Something broke here.

Sakurann:
superceded by IETF 115 and events

PR #181, opened 2022-11-24
revert to hash alg from digest derivation alg

Sakurann:
per IETF115 WG agreement.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -201,16 +201,13 @@ SD-JWT-DOC = (METADATA, SD-CLAIMS, NON-SD-CLAIMS)
 SD-JWT = SD-JWT-DOC | SIG(SD-JWT-DOC, ISSUER-PRIV-KEY)
 ```
 
-`SD-CLAIMS` is an object with claim names (`CLAIM-NAME`) mapped to the digests over the claim values (`CLAIM-VALUE`) with random salts (`SALT`). Digests are calculated using a digest derivation function such as a hash function, HMAC, or other (`DIGEST-DERIVATION()`):
+`SD-CLAIMS` is an object with claim names (`CLAIM-NAME`) mapped to the digests over the claim values (`CLAIM-VALUE`) with random salts (`SALT`). Digests are calculated using a hash function:
 
 ```
 SD-CLAIMS = (
-    DIGEST-DERIVATION(SALT, CLAIM-NAME, CLAIM-VALUE)
+    HASH(SALT, CLAIM-NAME, CLAIM-VALUE)
 )*
 ```
-
-When an HMAC or another type of derivation function is used for digest calculation, a secret cryptographic key or other cryptographic secret is used instead of a salt value.
-However, the term "salt" is used throughout this document for brevity.
 
 `SD-CLAIMS` can also be nested deeper to capture more complex objects, as will be shown later.
 
@@ -280,7 +277,7 @@ At a high level, the Verifier
 
  * receives the `COMBINED-PRESENTATION` from the Holder and verifies the signature of the SD-JWT using the Issuer's public key,
  * verifies the Holder Binding JWT, if Holder Binding is required by the Verifier's policy, using the public key included in the SD-JWT,
- * calculates the hash digests over the Holder-Selected Disclosures and verifies that each digest is contained in the SD-JWT.
+ * calculates the digest s over the Holder-Selected Disclosures and verifies that each digest is contained in the SD-JWT.
 
 The detailed algorithm is described in (#verifier_verification).
 
@@ -308,7 +305,7 @@ Usually, JSON-based formats transport claim values as simple properties of a JSO
 ```
 
 However, a problem arises when computation over the data need to be performed and verified, like signing or computing digests. Common signature schemes require the same byte string as input to the
-signature verification as was used for creating the signature. In the digest derivation approach outlined above, the same problem exists: for the Issuer and the
+signature verification as was used for creating the signature. In the digest approach outlined above, the same problem exists: for the Issuer and the
 Verifier to arrive at the same digest, the same byte string must be hashed.
 
 JSON [@!RFC7159], however, does not prescribe a unique encoding for data, but allows for variations in the encoded string. The data above, for example, can be encoded as
@@ -352,10 +349,10 @@ There are generally two approaches to deal with this problem:
    string for computing a digest.
 2. Source string hardening: Instead of transferring data in a format that
    may introduce variations, a representation of the data is serialized.
-   This representation is then used as the digest input at the Verifier,
+   This representation is then used as the hashing input at the Verifier,
    but also transferred to the Verifier and used for the same digest
-   calculcation there. This means that the Verifier can easily check the
-   digest over the byte string before finally deserializing and
+   calculcation there. This means that the Verifier can easily caompute and check the
+   digest of the byte string before finally deserializing and
    accessing the data.
 
 Mixed approaches are conceivable, i.e., transferring both the original JSON data
@@ -388,7 +385,7 @@ data. The original JSON data is then used by the application. See
 ## Format of an SD-JWT
 
 An SD-JWT is a JWT that MUST be signed using the Issuer's private key. The
-payload of an SD-JWT MUST contain the `sd_digest_derivation_alg` claim
+payload of an SD-JWT MUST contain the `sd_hash_alg` claim
 described in the following, MAY contain one or more selectively disclosable claims, and MAY contain a Holder's public key or a reference
 thereto, as well as further claims such as `iss`, `iat`, etc. as defined or
 required by the application using SD-JWTs.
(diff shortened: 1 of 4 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -280,7 +277,7 @@ At a high level, the Verifier
 
  * receives the `COMBINED-PRESENTATION` from the Holder and verifies the signature of the SD-JWT using the Issuer's public key,
  * verifies the Holder Binding JWT, if Holder Binding is required by the Verifier's policy, using the public key included in the SD-JWT,
- * calculates the hash digests over the Holder-Selected Disclosures and verifies that each digest is contained in the SD-JWT.
+ * calculates the digest s over the Holder-Selected Disclosures and verifies that each digest is contained in the SD-JWT.
```
danielfett, line 280:
```suggestion
 * calculates the digests over the Holder-Selected Disclosures and verifies that each digest is contained in the SD-JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -308,7 +305,7 @@ Usually, JSON-based formats transport claim values as simple properties of a JSO
 ```
 
 However, a problem arises when computation over the data need to be performed and verified, like signing or computing digests. Common signature schemes require the same byte string as input to the
-signature verification as was used for creating the signature. In the digest derivation approach outlined above, the same problem exists: for the Issuer and the
+signature verification as was used for creating the signature. In the digest approach outlined above, the same problem exists: for the Issuer and the
```
danielfett, line 308:
```suggestion
signature verification as was used for creating the signature. In the hash digest approach outlined above, the same problem exists: for the Issuer and the
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -352,10 +349,10 @@ There are generally two approaches to deal with this problem:
    string for computing a digest.
 2. Source string hardening: Instead of transferring data in a format that
    may introduce variations, a representation of the data is serialized.
-   This representation is then used as the digest input at the Verifier,
+   This representation is then used as the hashing input at the Verifier,
    but also transferred to the Verifier and used for the same digest
-   calculcation there. This means that the Verifier can easily check the
-   digest over the byte string before finally deserializing and
+   calculcation there. This means that the Verifier can easily caompute and check the
```
danielfett, line 354:
```suggestion
   calculcation there. This means that the Verifier can easily compute and check the
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -428,24 +425,24 @@ are all valid and encode the same claim value:
...
+For embedding the Disclosures in the SD-JWT, the Disclosures are hashed using the hash algorithm specified in the `sd_hash_alg` claim described below. The resulting digest is then included in the SD-JWT instead of the original claim value, as described next.
 
-The hash digest MUST be taken over the US-ASCII bytes of the base64url-encoded Disclosure. This follows the convention in JWS [@RFC7515] and JWE [@RFC7516]. The bytes of the hash digest MUST then be base64url-encoded.
+The digest MUST be taken over the US-ASCII bytes of the base64url-encoded Disclosure. This follows the convention in JWS [@RFC7515] and JWE [@RFC7516]. The bytes of the digest MUST then be base64url-encoded.
 
 It is important to note that:
 
  * The input to the hash function is the base64url-encoded Disclosure, not the bytes encoded by the base64url string.
- * The bytes of the output of the hash function are base64url-encoded, not the bytes making up the (often used) hex representation of the bytes of the hash digest.
+ * The bytes of the output of the hash function are base64url-encoded, not the bytes making up the (often used) hex representation of the bytes of the digest .
```
danielfett, line 435:
```suggestion
 * The bytes of the output of the hash function are base64url-encoded, not the bytes making up the (often used) hex representation of the bytes of the digest.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -428,24 +425,24 @@ are all valid and encode the same claim value:
...
+SHA-256 digest of the Disclosure `WyI2cU1RdlJMNWhhaiIsICJmYW1pbHlfbmFtZSIsICJNw7ZiaXVzIl0` would be
 `uutlBuYeMDyjLLTpf6Jxi7yNkEF35jdyWMn9U7b_RYY`.
 
 #### Decoy Digests {#decoy_digests}
 
-An Issuer MAY add additional hash digests to the SD-JWT that are not associated with any claim.  The purpose of such "decoy" digests is to make it more difficult for an attacker to see the original number of claims contained in the SD-JWT. It is RECOMMENDED to create the decoy digests by hashing over a cryptographically secure random number. The bytes of the hash digest MUST then be base64url-encoded as above. The same digest function as for the Disclosures MUST be used.
+An Issuer MAY add additional digest s to the SD-JWT that are not associated with any claim.  The purpose of such "decoy" digests is to make it more difficult for an attacker to see the original number of claims contained in the SD-JWT. It is RECOMMENDED to create the decoy digests by hashing over a cryptographically secure random number. The bytes of the digest MUST then be base64url-encoded as above. The same digest function as for the Disclosures MUST be used.
```
danielfett, line 443: Please fix `digest s` → `digests` everywhere.

danielfett: Why replace `hash digests` with `digests` everywhere?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -689,11 +684,11 @@ To this end, Verifiers MUST follow the following steps (or equivalent):
     2. Validate the signature over the SD-JWT.
     3. Validate the Issuer of the SD-JWT and that the signing key belongs to this Issuer.
     4. Check that the SD-JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the SD-JWT, and not selectively disclosed.
-    5. Check that the `sd_digest_derivation_alg` claim is present and its value is understood and the digest derivation algorithm is deemed secure.
+    5. Check that the `sd_hash_alg` claim is present and its value is understood and the hash algorithm is deemed secure.
  4. Create a copy of the SD-JWT payload, if required for further processing.
  5. Process the Disclosures. For each Disclosure provided:
-    1. Calculate the hash digest over the base64url string as described in (#hashing_disclosures).
-    2. Find all `_sd` keys in the SD-JWT payload that contain a hash digest calculated in the previous step. Note that there might be more than one `_sd` arrays in on SD-JWT.
+    1. Calculate the digest  over the base64url string as described in (#hashing_disclosures).
```
danielfett, line 690:
```suggestion
    1. Calculate the digest over the base64url string as described in (#hashing_disclosures).
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -689,11 +684,11 @@ To this end, Verifiers MUST follow the following steps (or equivalent):
...
+    2. Find all `_sd` keys in the SD-JWT payload that contain a digest  calculated in the previous step. Note that there might be more than one `_sd` arrays in on SD-JWT.
```
danielfett, line 691:
```suggestion
    2. Find all `_sd` keys in the SD-JWT payload that contain a digest calculated in the previous step. Note that there might be more than one `_sd` arrays in on SD-JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -768,24 +763,26 @@ guess. A new salt MUST be chosen for each claim.
 
 The RECOMMENDED minimum length of the randomly-generated portion of the salt is 128 bits.
 
-Note that minimum 128 bits would be necessary when SHA-256, HMAC-SHA256, or a function of similar strength is used, but a smaller salt size might achieve similar level of security if a stronger iterative derivation function is used.
+Note that minimum 128 bits would be necessary when SHA-256, or a function of similar strength is used.
```
danielfett, line 766: I don't think this sentence adds information, the RECOMMENDATION is already there.
```suggestion
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -768,24 +763,26 @@ guess. A new salt MUST be chosen for each claim.
...
-For the security of this scheme, the digest derivation algorithm is required to be preimage and collision
+For the security of this scheme, the hash algorithm is required to be preimage and collision
 resistant, i.e., it is infeasible to calculate the salt and claim value that result in
 a particular digest, and it is infeasible to find a different salt and claim value pair that
 result in a matching digest, respectively.
 
 Furthermore the hash algorithms MD2, MD4, MD5, RIPEMD-160, and SHA-1
 revealed fundamental weaknesses and they MUST NOT be used.
 
+Note that implementations willing to use functions other than hash functions, for example to achieve a smaller salt size with similar level of security by using a stronger iterative derivation function, are RECOMMENDED to define a profile of this specification, i.e., when an HMAC or another type of hash function is used for digest calculation, it needs to be defiend how a secret cryptographic key or other cryptographic secret is can be used instead of a salt value.
```
danielfett, line 784: Technically speaking, that would not be a profile, as it would need to change normative things within this spec. We cannot really give this recommendation here. All we can say is: HMAC or similar is at your own risk, non-standard, etc.; I'm not sure if we want to say that here.

danielfett (changes requested):

danielfett:
Can you please recompute the examples?

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi, line 280: lots of `digest s` that need to be fixed to `digests`

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I think what we are trying to say here is that a smaller salt could potentially be used in conjunction with a more computationally expensive hash function. Where a hash function is a single input -> single digest output. And the  Hash Function Claim section says sd_hash_alg can be "a value defined in another specification and/or profile of this specification". Such a spec/profile would need to define that hash function for use at the SD-JWT layer as a single input to single digest output thing. But it could map/profile something like PBKDF2 to do that or define a 30 billion iteration sha2 thing. 

I'm not sure how much of that we want to say here. But it would be an "at your own risk" kind of thing. 

I think that the specific talk of HMAC and cryptographic secrets here is confusing things though. It certainly confused me :)

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -388,7 +385,7 @@ data. The original JSON data is then used by the application. See
 ## Format of an SD-JWT
 
 An SD-JWT is a JWT that MUST be signed using the Issuer's private key. The
-payload of an SD-JWT MUST contain the `sd_digest_derivation_alg` claim
+payload of an SD-JWT MUST contain the `sd_hash_alg` claim
```
bc-pi, line 388: I'd like to be consistent with the sd names.  So either 

`_sd` and `_sd_hash_alg` 

or

`sd` and `sd_hash_alg`

But not mixed.

danielfett: +1 for `_sd_hash_alg`. 

BTW, I added the underscore in front of `_sd` because it sorts nicely to the top (in case the claims are sorted alphabetically by key). With just `sd` the hashes sit somewhere in between the other non-sd claims.

danielfett: `_sd_alg` may work as well.

bc-pi: `_sd_alg` would work for me

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: fixed

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I think `_sd_hash_alg` is explicit and nice?

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: > Why replace hash digests with digests everywhere?

because digest implies it is a hash digest?

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: > the Hash Function Claim section says sd_hash_alg can be "a value defined in another specification and/or profile of this specification"
personally, I am ok removing this section, with this PR, don't think it's relevant anymore.

Sakurann: also removed this clause to prevent confusion.

Sakurann: cc @christianpaquin for transparency that during IETF115 OAuth WG, there was a decision to revert to hashing alg because including HMAC includes a lot of additional definitions to be made, not just saying _sd_hash_alg can be HMAC, and we should keep SD-JWT spec simple.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -768,24 +763,26 @@ guess. A new salt MUST be chosen for each claim.
...
-## Choice of a digest derivation algorithm
+## Choice of a Hash Algorithm
 
-For the security of this scheme, the digest derivation algorithm is required to be preimage and collision
+For the security of this scheme, the hash algorithm is required to be preimage and collision
 resistant, i.e., it is infeasible to calculate the salt and claim value that result in
 a particular digest, and it is infeasible to find a different salt and claim value pair that
 result in a matching digest, respectively.
 
 Furthermore the hash algorithms MD2, MD4, MD5, RIPEMD-160, and SHA-1
```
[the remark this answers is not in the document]
Sakurann: externalized to Issue #183 so that we can merge this PR now, publish and come back to this later

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I'd prefer `_sd_alg`. Size isn't everything but keeping JWTs as small as reasonable is worth keeping in mind. And JWT claim names are typically rather short/concise. `_sd_alg` is shorter but still conveys enough about what it is.

danielfett (approved):

bc-pi (approved):

PR #184, opened 2022-12-06
First draft for enveloping considerations

danielfett:


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -841,6 +867,25 @@ the SD-JWT.
...
+Usually, transport protocols for issuance and presentation of credentials
+are designed to protect the confidentiality of the transmitted data, for
+example, by requiring the use of TLS.
+
+This specification therefore considers the confidentiality of the data to be
+provided by the transport protocol and does not specify any encryption
+mechanism.
+
+Implementers MUST ensure that the transport protocol provides confidentiality
+if the privacy of End-User data or correllation attacks are a concern. Implementers MAY define a
```
bc-pi, line 885:
```suggestion
if the privacy of End-User data or correlation attacks are a concern. Implementers MAY define a
```

bc-pi: It's not a competition :)

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -841,6 +867,25 @@ the SD-JWT.
...
+example, by requiring the use of TLS.
+
+This specification therefore considers the confidentiality of the data to be
+provided by the transport protocol and does not specify any encryption
+mechanism.
+
+Implementers MUST ensure that the transport protocol provides confidentiality
+if the privacy of End-User data or correlation attacks are a concern. Implementers MAY define a
+JWT envelope format (as described in (#enveloping)) to encrypt the SD-JWT
+and associated Disclosures when transmitted over an insecure channel.
```
bc-pi, line 887:
```suggestion
Implementers MUST ensure that the transport protocol provides confidentiality,
if the privacy of End-User data or correlation attacks are a concern. Implementers MAY define an
envelope format (such as described in (#enveloping) or nesting the SD-JWT Combined Format as
the plaintext of a JWE) to encrypt the SD-JWT
and associated Disclosures when transmitted over an insecure channel
```

danielfett: Should 'plaintext' be 'payload'? Other than that, looks good!

bc-pi: as mentioned on one of the messaging platforms - I think nesting the SD-JWT and Disclosures directly in a JWE is an appropriate approach in some/many cases. This suggestion just tries to leave room for that.

bc-pi: > Should 'plaintext' be 'payload'? Other than that, looks good!

AFAIK JWE/T uses 'plaintext' and 'payload' somewhat interchangeably and sometimes together.  I updated the suggestion to have both "plaintext payload".

danielfett: I see it says 'plaintext payload' now. Thanks, I'll merge this!

bc-pi (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -841,6 +867,26 @@ the SD-JWT.
 
 # Privacy Considerations {#privacy_considerations}
 
+## Confidentiality during Transport
+
+If the SD-JWT and associated Disclosures are transmitted over an insecure
+channel during issuance or presentation, an adversary may be able to
+intercept and read the End-User's personal data or correlate the information with previous uses of the same SD-JWT.
```
Sakurann, line 874:
```suggestion
intercept and read the End-User's personal data.
```
How does the insecure channel related to correlating the multiple usage of the same SD-JWT..? isn't it more of an identifier/signing key not being pairwise issue..?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -841,6 +867,26 @@ the SD-JWT.
...
+
+This specification therefore considers the confidentiality of the data to be
+provided by the transport protocol and does not specify any encryption
+mechanism.
+
+Implementers MUST ensure that the transport protocol provides confidentiality,
+if the privacy of End-User data or correlation attacks are a concern. Implementers MAY define an
+envelope format (such as described in (#enveloping) or nesting the SD-JWT Combined Format as
+the plaintext payload of a JWE) to encrypt the SD-JWT
+and associated Disclosures when transmitted over an insecure channel.
```
Sakurann, line 888:
```suggestion
If the privacy of End-User data or correlation attacks are a concern, 
Implementers MUST ensure that the transport protocol provides confidentiality. 
When transmitted over an insecure channel, implementers MAY also define an
envelope format (such as described in (#enveloping) or nesting the SD-JWT Combined Format as
the plaintext payload of a JWE) to encrypt the SD-JWT
and associated Disclosures.
```

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: The solution to both problems is the same, but 

1. colluding verifiers identifying a person using the same signing keys, and
2. attackers observing the wire and seeing the same credential

are two different problems.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: This changes the meaning (in an unintended way). Implementers may define an envelope format even for secure channels. 

(Also, the second sentence may imply that Implementers are sometimes transmitted over insecure channels... )

Sakurann (approved):
made comments/suggestions, but overall sounds good.

PR #190, opened 2022-12-07
Longer salt

bc-pi:
Longer salt, which meets the recommendations, in examples in Creating… Disclosures

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -408,18 +408,18 @@ The following example illustrates the steps described above.
 
 The array is created as follows:
 ```json
-["6qMQvRL5haj", "family_name", "Möbius"]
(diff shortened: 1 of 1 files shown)


bc-pi:
While working on this PR, I was going to a base64 decoding site to do a bit of a sanity check on the values but got distracted and decided to see how ChatGPT would handle it. So I gave it this prompt in the openai playground:

```decode WyJfMjZiYzRMVC1hYzZxMktJNmNCVzVlcyIsICJmYW1pbHlfbmFtZSIsICJNw7ZiaXVzIl0``` 

and it added one `=` pad character to the end of the prompt and answered with:

```["f26bc4L-ac6q2KI6cBW5es", "family_name", "Müller"]```

Note that it actually decodes to:

```
["_26bc4LT-ac6q2KI6cBW5es", "family_name", "Möbius"]
```

if you need me, I'll just be over here staring into the void.

bc-pi:
It's amazing and terrifying at the same time.

PR #192, opened 2022-12-08
Use ES256 instead of RS256 to generate shorter examples

danielfett:


sd_jwt/operations.py
```diff
@@ -231,7 +233,7 @@ def _create_signed_jwt(self):
 
         # Sign the SD-JWT using the issuer's key
         self.sd_jwt = JWS(payload=dumps(self.sd_jwt_payload))
-        _headers = {"alg": self._sign_alg, "kid": self._issuer_key.thumbprint()}
+        _headers = {"alg": self._sign_alg}
```
danielfett, line 236: @bc-pi I removed the "kid" header parameter from the examples. It used to contain the key thumbprint. AFAIK there is no requirement for a "kid" header parameter and just always putting the thumbprint there is not the right solution anyway - any objections?

danielfett: (I'm not sure why `kid` was in there in the first place.)

peppelinux: It was me @danielfett, it's not relevant for the draft and its examples, please go ahead :-)

bc-pi: no objections

examples/simple/sd_jwt_payload.json
```diff
@@ -14,14 +14,13 @@
...
-        659kg4hVo9dbGoCJE3ZGF_eaetE30UhBUEgpGwrDrQiJ9zqprmcFfr3qvvkG
-        jtth8Zgl1eM2bJcOwE7PCBHWTKWYs152R7g6Jg2OVph-a8rq-q79MhKG5QoW
-        _mTz10QT_6H4c7PjWG1fjh8hpWNnbP_pv6d1zSwZfc5fl6yVRL0DV0V3lGHK
-        e2Wqf_eNGjBrBLVklDTk8-stX_MWLcR-EGmXAOv0UBWitS_dXJKJu-vXJyw1
-        4nHSGuxTIK2hx1pttMft9CsvqimXKeDTU14qQL1eE7ihcw",
-      "e": "AQAB"
+      "kty": "EC",
+      "use": "sig",
+      "alg": "ES256",
+      "kid": "2",
```
bc-pi, line 20: I don't think `use`, `alg` and `kid` should be included here.

examples/simple/verified_contents.json
```diff
@@ -5,14 +5,13 @@
...
-        659kg4hVo9dbGoCJE3ZGF_eaetE30UhBUEgpGwrDrQiJ9zqprmcFfr3qvvkG
-        jtth8Zgl1eM2bJcOwE7PCBHWTKWYs152R7g6Jg2OVph-a8rq-q79MhKG5QoW
-        _mTz10QT_6H4c7PjWG1fjh8hpWNnbP_pv6d1zSwZfc5fl6yVRL0DV0V3lGHK
-        e2Wqf_eNGjBrBLVklDTk8-stX_MWLcR-EGmXAOv0UBWitS_dXJKJu-vXJyw1
-        4nHSGuxTIK2hx1pttMft9CsvqimXKeDTU14qQL1eE7ihcw",
-      "e": "AQAB"
+      "kty": "EC",
+      "use": "sig",
+      "alg": "ES256",
+      "kid": "2",
```
bc-pi, line 11: not really meaningful or useful in this context (I don't think?) but take up space and potentially distract or confuse

sd_jwt/demo_settings.py
```diff
@@ -1,31 +1,27 @@
...
 
+HOLDER_KEY_DICT = {
+    "kty": "EC",
+    "d": "5K5SCos8zf9zRemGGUl6yfok-_NiiryNZsvANWMhF-I",
+    "use": "sig",
+    "crv": "P-256",
+    "kid": "2",
+    "x": "TCAER19Zvu3OHF4j4W4vfSVoHIP1ILilDls7vCeGemc",
+    "y": "ZxjiWWbZMQGHVWKVQ4hbSIirsVfuecCE6t4jT9F2HZQ",
+    "alg": "ES256"
```
bc-pi, line 26: I'm gussing this is where they come from but, as per my previous comments, I don't think use/kid/alg should be in the holder confirmation key.

examples/simple/verified_contents.json
bc-pi: use alg and kid

bc-pi (changes requested):
Looks pretty good overall! I even double checked some of the signatures with my JOSE/JWT library. 

I do, however, think that before merging a few parameters should be removed from the holder jwk.

examples/simple/verified_contents.json
danielfett: You're right! I removed them.

sd_jwt/demo_settings.py
danielfett: As above - removed the three superfluous elements.

bc-pi (approved):

bc-pi:
Approved!

Sakurann (approved):

PR #193, opened 2022-12-23
Examples for eIDAS pid, eaa, and JSON-LD credential

tlodderstedt:
Hi, 

this PR adds examples for PID and EAA (Diploma based on EBSI Schema) as SD-JWT-VCs and an example of a SD-JWT with JSON-LD payload (Vaccination Certificate).


examples/a/disclosures.md
```diff
@@ -0,0 +1,128 @@
...
+["did:ebsi:zsSgDXeYPhZ3AuKhTFneDf1"], "specifiedBy": {"id":
+"urn:epass:asssessmentspec:1", "title": "Overall Diploma
+Assessment", "gradingScheme": {"id":
+"urn:epass:scoringschemespec:1", "title": "General grading scheme in
+Croatia", "definition": "The Croatian national grading scheme
+consists of five grades with numerical equivalents: izvrstan \u2013
+5 (outstanding); vrlo dobar \u2013 4 (very good); dobar \u2013 3
+(good); dovoljan \u2013 2 (sufficient); nedovoljan \u2013 1
+(insufficient - fail). The minimum passing grade is dovoljan \u2013
+2."}}}]}]]
```
peppelinux, line 74: I suggest an indented representation of this JSON for better readability

danielfett: This is auto generated by the script and I suppose there's no need to include this markdown snippet in the spec itself.

bc-pi:
> to clarify, the idea is not to add this to the main spec text, but purely add examples in this repo for people who want to generate them?

I think it'd be more appropriate to do that kind of thing in a fork of the repo and keep the scope of the main repo to the SD-JWT document itself.

PR #202, opened 2023-01-31
Independent and Secret Salts

danielfett:
Makes these two points in John Mattsson's review explicit:

> - The salts need to be secret. Otherwise, an attacker can guess and verify claims.
> - The salts need to be independent of each other.  Otherwise, a Verifier can guess claims.


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -607,7 +607,8 @@ as follows (line break added for readability):
 This is called the Combined Format for Presentation.
 
 The Holder MAY send any subset of the Disclosures to the Verifier, i.e.,
-none, multiple, or all Disclosures.
+none, multiple, or all Disclosures. The Holder MUST NOT send Disclosures
+or salt values for data that should not be revealed to the Verifier.
```
Sakurann, line 611: salt is part of the disclosures, so maybe something like:
```suggestion
none, multiple, or all Disclosures. The Holder MUST NOT send Disclosures,
especiallt salt values, for data that should not be revealed to the Verifier.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -403,7 +403,7 @@ For each claim that is to be selectively disclosed, the Issuer creates a Disclos
 The Issuer MUST create a Disclosure for each selectively disclosable claim as follows:
 
  * Create an array of three elements in this order:
-   1. A salt value. See (#salt-entropy) and (#salt_minlength) for security considerations. The salt value MUST be unique for each claim that is to be selectively disclosed. It is RECOMMENDED to base64url-encode the salt value, producing a string. Any other type that is allowed in JSON MAY be used, e.g., a number.
+   1. A salt value. See (#salt-entropy) and (#salt_minlength) for security considerations. The salt value MUST be unique for each claim that is to be selectively disclosed. It is RECOMMENDED to base64url-encode the salt value, producing a string. Any other type that is allowed in JSON MAY be used, e.g., a number. The salt MUST NOT be disclosed to third parties except the Holder.
```
Sakurann, line 406: do we need a statement saying that the holder/wallet must protect Disclosures, and in particular, salts with particular care? I mean, the spec of disclosing them is not the only place when salts can get to the hands on an attacker..

Sakurann, line 406: probably it is this sentence that would benefit from clarifying?: "The salt value MUST be unique for each claim that is to be selectively disclosed."
to something like "A new salt MUST be chosen for each claim independently from other salts." in a change below?

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: "especially" makes it sound like the salt is somehow more sensitive than the claim name/value. Which is not the case at all. The salt only adds randomness to the hash input  to make it infeasible to "reverse" the hash value by enumerating potential claim name/values into the hash function. The salt does nothing from the privacy perspective, if the claim is reveled.

bc-pi: maybe something like this, which acknowledges that salt is part of the disclosure but doesn't over emphasize it: 
```suggestion
none, multiple, or all Disclosures. The Holder MUST NOT send Disclosures
(including salt values) for data that should not be revealed to the Verifier.
```
?

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: Being pedantic "MUST NOT be disclosed to third parties except the Holder" suggests that it can't be sent to any Verifier. I do realize this is in the context of the Issuer creating the thing but can see easy confusion.

bc-pi: maybe something like this? 

```suggestion
   1. A salt value. See (#salt-entropy) and (#salt_minlength) for security considerations. The salt value MUST be unique for each claim that is to be selectively disclosed. It is RECOMMENDED to base64url-encode the salt value, producing a string. Any other type that is allowed in JSON MAY be used, e.g., a number. The Issuer MUST NOT disclose the salt value to third parties except the Holder.
```

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: I wanted to say something like "don't reveal the disclosures, but also don't take the salt values out of them and reveal them separately". 

Also I just notice that I don't like the 'should' in the original sentence.

```suggestion
none, multiple, or all Disclosures. For data that the Holder does not want to reveal
to the Verifier, the Holder MUST NOT send Disclosures or reveal the salt values in any
other way.
```

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: Would you keep the other sentence that I added further below?

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I can live with that.

Sakurann: me too

bc-pi (approved):

Sakurann (approved):

Issue #175, opened 2022-11-22
Location of "Challenge of Canonicalization Section"


Sakurann:
Suggest we move this section lower, as with the new encoding approach I anticipate less questions regarding canonicalization, while this is a good clarification.

https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/174/files#r1028969018

bc-pi:
I recently had a conversation with a guy (editor of RFC7520 FWIW) who had some comments/observations that would support the idea of moving it into an appendix.

"my opinion: the entire discussion about canonicalization only serves to distract.  This document should only ever talk about the source hardening; if you really want/need to talk about c14n, put it in an appendix at the end" and "then that entire Section 5.1 ought to just be dropped then, I think"

Sakurann:
ok, let's move to appendix

PR #205, opened 2023-02-03
move (& rename) the c14n challenges section to an appendix

bc-pi:
Rename and move the c14n section to an appendix. Also add a paragraph attempting to provide a bit more context in the new location. 

for #175

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -50,7 +50,7 @@ documents that support selective disclosure of JWT claims.
 
 # Introduction {#Introduction}
 
-The JSON-based representation of claims in a signed JSON Web Token (JWT) [@!RFC7519] is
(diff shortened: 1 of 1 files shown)


Sakurann (approved):
I wasn't totally sure about the title "Disclosure Format Considerations", but can't think of a better one and after re-reading the content, I think it captures the goal.

bc-pi:
I'm not thrilled with the section name either but thought "The Challenge of Canonicalization" felt really out of place in an appendix so wanted to change it. "Disclosure Format Considerations" was the best I could come up with.

danielfett:
Looks good to me, but we should probably reference the new appendix somewhere in the main body.

bc-pi:
> Looks good to me, but we should probably reference the new appendix somewhere in the main body.

Please add such a reference wherever you think is appropriate.

bc-pi:
> Looks good to me, but we should probably reference the new appendix somewhere in the main body. 

went ahead and did that with 0db6e9865fedf8437e75b9f2eb9d12fc52ee48f1

danielfett:
Thank you!

PR #206, opened 2023-02-03
no MACs

bc-pi:
More explicitly state that SD-JWTs have to be signed asymmetrically (no MAC and no `none`)

One item out of John Mattsson's review: https://mailarchive.ietf.org/arch/msg/oauth/QhyXsK8ZXFBbkIcH5LNjMPC1zX4/

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -454,6 +454,8 @@ To ensure readability and replicability, the examples in this specification do n
 #### Creating an SD-JWT {#creating_sd_jwt}
 
 An SD-JWT is a JWT that MUST be signed using the Issuer's private key.
+It MUST use a JWS asymmetric digital signature algorithm from [@IANA.JWS.Algorithms].
+It MUST NOT use `none` or an identifier for a symmetric algorithm (MAC).
 
 An SD-JWT MAY contain both selectively disclosable claims and non-selectively disclosable claims, i.e., claims that are always contained in the SD-JWT in plaintext and are always visible to a Verifier.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -454,6 +454,8 @@ To ensure readability and replicability, the examples in this specification do n
 #### Creating an SD-JWT {#creating_sd_jwt}
 
 An SD-JWT is a JWT that MUST be signed using the Issuer's private key.
+It MUST use a JWS asymmetric digital signature algorithm from [@IANA.JWS.Algorithms].
```
Sakurann, line 457: Are there any asymmetric digital signatures that are not registered in IANA? Are we prohibiting those?

bc-pi: That's a good question. There might be unregistered ones (JWS allows for it https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.1) and this text would prohibit their use in SD-JWT. Which is kind of an edge case but probably not what we want. I borrowed that text from DPoP where an AD had suggested it be more restrictive with respect to the registry. But I didn't love it there and your question reminds me of that and that it's probably less appropriate here. I'll rephrase this sentence so as not to be prohibitive.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -454,6 +454,8 @@ To ensure readability and replicability, the examples in this specification do n
...
+It MUST NOT use `none` or an identifier for a symmetric algorithm (MAC).
```
bc-pi, line 458:
```suggestion
It MUST use a JWS asymmetric digital signature algorithm and
MUST NOT use `none` or an identifier for a symmetric algorithm (MAC).
```

danielfett (approved):

Issue #183, opened 2022-12-02
truncated hashes


Sakurann:
> Do we need to say anything about truncated hashes here? I think probably so. The [registry](https://www.iana.org/assignments/named-information/named-information.xhtml) has several including sha-256-128, which is probably okay security-wise and even useful as a space saving device. But it also has sha-256-32, which is probably never okay in this context. And a few in between those two that I'm not sure about.

from this PR https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/181#pullrequestreview-1195712714

PR #207, opened 2023-02-06
a little more on Choice of a Hash Algorithm 

bc-pi:
A bit more in security considerations for Choice of a Hash Algorithm (1st & 2nd preimage resistant and not majorly truncated) for issue  #183

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -805,12 +805,16 @@ a different salt.
 
 ## Choice of a Hash Algorithm
 
-For the security of this scheme, the hash algorithm is required to be preimage and collision
+For the security of this scheme, the hash algorithm is required to be first and second preimage
 resistant, i.e., it is infeasible to calculate the salt and claim value that result in
 a particular digest, and it is infeasible to find a different salt and claim value pair that
-result in a matching digest, respectively.
+result in a matching digest, respectively. Hash algorithms that do not meet the aforementioned
+requirements MUST NOT be used. Note that inclusion in the "Named Information Hash Algorithm"
+registry [@IANA.Hash.Algorithms] alone does not indicate a hash algorithm's suitability for use
+in SD-JWT (it contains several heavily truncated digests, such as
+`sha-256-32` and `sha-256-64`, which are unfit for security applications).
 
-Furthermore the hash algorithms MD2, MD4, MD5, RIPEMD-160, and SHA-1
+Furthermore, the hash algorithms MD2, MD4, MD5, RIPEMD-160, and SHA-1
 revealed fundamental weaknesses and they MUST NOT be used.
 
 ## Holder Binding {#holder_binding_security}
@@ -1221,6 +1225,7 @@ Disclosures:
    * Discussion on holder binding and privacy of stored credentials
    * Add some context about SD-JWT being general-purpose despite being a product of the OAuth WG
    * Use ES256 instead of RS256 in examples
+   * A bit more in security considerations for Choice of a Hash Algorithm (1st & 2nd preimage resistant and not majorly truncated)
    * Fix the Document History (which had a premature list for -03)
 
    -02


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -805,12 +805,16 @@ a different salt.
...
-For the security of this scheme, the hash algorithm is required to be preimage and collision
+For the security of this scheme, the hash algorithm is required to be first and second preimage
 resistant, i.e., it is infeasible to calculate the salt and claim value that result in
 a particular digest, and it is infeasible to find a different salt and claim value pair that
-result in a matching digest, respectively.
+result in a matching digest, respectively. Hash algorithms that do not meet the aforementioned
+requirements MUST NOT be used. Note that inclusion in the "Named Information Hash Algorithm"
+registry [@IANA.Hash.Algorithms] alone does not indicate a hash algorithm's suitability for use
+in SD-JWT (it contains several heavily truncated digests, such as
+`sha-256-32` and `sha-256-64`, which are unfit for security applications).
```
Sakurann, line 815:
```suggestion
result in a matching digest, respectively. 

Hash algorithms that do not meet the aforementioned requirements MUST NOT be used. 
Inclusion in the "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] 
alone does not indicate a hash algorithm's suitability for use in SD-JWT (it contains several
heavily truncated digests, such as `sha-256-32` and `sha-256-64`, which are unfit for security
applications).
```

Sakurann (approved):

bc-pi:
While looking at this I noticed https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-02.html#name-hash-function-claim only allows hash algs from the registry.  https://github.com/oauth-wg/oauth-selective-disclosure-jwt/commit/b836ab1e9f8001221ed5245d75faa90680bc86d6 removed the JOSE HMAC stuff but also removed allowing the hash alg to be "a value defined in another specification and/or profile of this specification" which I thought we wanted to have - for Christian's type use cases and extensibility in general. 

This isn't the right place for a discussion but wanted to note it somewhere before our call.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -805,12 +805,18 @@ a different salt.
 
 ## Choice of a Hash Algorithm
 
-For the security of this scheme, the hash algorithm is required to be preimage and collision
+For the security of this scheme, the hash algorithm is required to be first and second preimage
 resistant, i.e., it is infeasible to calculate the salt and claim value that result in
 a particular digest, and it is infeasible to find a different salt and claim value pair that
 result in a matching digest, respectively.
```
danielfett, line 811:
```suggestion
a particular digest, and, for any salt and claim value pair, it is infeasible to find a different salt and claim value pair that
result in the same digest, respectively.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -805,12 +805,18 @@ a different salt.
 
 ## Choice of a Hash Algorithm
 
-For the security of this scheme, the hash algorithm is required to be preimage and collision
+For the security of this scheme, the hash algorithm is required to be first and second preimage
```
danielfett, line 808: Is first and second preimage resistance a common term? I usually see "preimage resistant" and "second-preimage resistant".

danielfett (changes requested):

Sakurann:
I think we intentionally removed “a value defined in another specification and/or profile of this specification” Because using that would require a separate profile and not just defining an identifier.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi:
```suggestion
For the security of this scheme, the hash algorithm is required to be preimage resistant and second-preimage 
```

PR #209, opened 2023-02-09
Tidy up example in Appendix 4

bc-pi:
basically removing the example JWT header JSON (which aren't shown elsewhere FWIW) that had a fictitious `typ` and meaningless `kid` and didn't add anything to understanding/following the example


Sakurann (approved):
we may want to align the language around:

"The SD-JWT payload is as follows:"
vs
"The following shows the resulting SD-JWT payload:"
etc.

but agree in principle.

danielfett (approved):

PR #210, opened 2023-02-10
Example PID

tlodderstedt:
Here is an example of a PID (Person Identification Data) (see https://digital-strategy.ec.europa.eu/en/library/european-digital-identity-wallet-architecture-and-reference-framework)

This example has an array called `binding`, where every object contains the data for a certain holder binding method (cryptographic and biometric). My assumption is that a holder would only disclose the data for one method with a certain verifier, so this example also illustrates the need to be able to selectively disclose individual objects from an array.


Sakurann:
the formatting of the yml file has changed a bit. I can do a PR conformant to a new ver. suggest merging after that to have these examples in the repo even if they are not used in the actual spec text.

Sakurann:
closing in favor of #304

PR #211, opened 2023-02-10
Example EAA 

tlodderstedt:
This is an example of a "Electronic Attestation of Attributes" (see https://digital-strategy.ec.europa.eu/en/library/european-digital-identity-wallet-architecture-and-reference-framework). 

The example if a diploma. The schema is based on the EBSi schema for diplomas (but uses plain JSON). The credential uses claims based holder binding, i.e. it contains a declaration of the name and date of birth of the legitimate holder. The holder needs to prove this claims separately from the diploma. She can, for example, use the PID from PR #210.


Sakurann:
the formatting of the yml file has changed a bit. I can do a PR conformant to a new ver. suggest merging after that to have these examples in the repo even if they are not used in the actual spec text.

Sakurann:
moving to sd-jwt-vc repo

Issue #87, opened 2022-06-26
Security and privacy targets not defined


sakimura:
# Comments
From the current version of the document, the security and privacy target is not really defined but only a mechanism is defined as “An SD-JWT, at its core, is a digitally signed document containing hashes over the claim values with unique salts and other metadata. It MUST be digitally signed using the issuer's private key”. Explicitly defining security and privacy goals may be desired. 

# Proposal
 
Add the following at the beginning of “Concept”

From the point of view of the collection limitation, which is one of the main privacy principles, it is desirable to be able to provide only the claims necessary for the verifier to process. In the case of a standard JWT [RFC7519], however, this can only be achieved when the JWT is minted afresh each time and it is not possible to re-use the JWT that is pre-minted for other occasions. For example, a holder may want to re-use the document that contains name, address (municipality, street address, room number), date of birth, and National Identification Number to prove that he is a resident of the municipality so that he can receive the service to the resident. In the case of the normal JWT, this is not possible as the signature is over the entire claim set directly. To achieve it, a JWT that only contains the claim about the municipality needs to be minted. 

Selective disclosure JWT (SD-JWT) aims to address this issue by introducing an indirection: sign the salted hash of the claims instead of the claim set and provide only the claim value of the claim that is required to be processed by the receiver, which assumes the verifier role. This way, while the claim names and the salted hash of the claim values of all the claims contained in the document issued by the issuer is disclosed, the claim values of other claims are not in principle. 

In this document, a protocol is considered to fulfil the selective disclosure property when it has the following properties: 

## R1. Strong uncoverability

A protocol is said to achieve strong uncoverability if the adversary acting as the verifier cannot successfully find the claim values that were not intended to be disclosed to the verifier, with a probability significantly better than negligible. 

## R2. Weak uncoverability

While strong uncoverability is desirable, it is not trivial to achieve it. For example, if the original document was a passport that came with the signature of the national authority, then just by looking at the issuer and its signature, it is trivial to find the nationality of the subject in the release. Also, if the existence of the claim name in the release makes it possible for the verifier to guess the value with a probability significantly better than negligible, then it does not satisfy strong uncoverability. 

With these considerations, weak uncoverability is defined as follows: 

A protocol is said to achieve weak uncoverability if the adversary acting as the verifier cannot successfully find the claim values that were not intended to be disclosed to the verifier from the section of the release that directly corresponds with the claim, with a probability significantly better than negligible.  

[...]

PR #92, opened 2022-06-26
To address issue #87 Security and privacy targets not defined

sakimura:
To address issue #87

PR #225, opened 2023-03-06
security targets

Sakurann:
This PR outlines security targets of SD-JWT.
when merged it will replace PR #92 and close issue #87.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -599,7 +599,7 @@ Here, `eyJhbGci...emhlaUJhZzBZ` represents the SD-JWT and `eyJhb...dYALCGg` repr
 
 # Security Considerations {#security_considerations}
 
-Security considerations in this section help achieve the following the following properties:
+Security considerations in this section help achieve the following properties:
 
 * Claim values issued by the Issuer cannot be obtained by the adversary,
 	- if the Holder did not intend to disclose them,
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -599,6 +599,14 @@ Here, `eyJhbGci...emhlaUJhZzBZ` represents the SD-JWT and `eyJhb...dYALCGg` repr
 
 # Security Considerations {#security_considerations}
 
+Security considerations in this section help achieve the following the following properties:
+
+* Claim values issued by the Issuer cannot be obtained by the adversary,
+	- if the Holder did not intend to disclose them,
+	- even if the content of the protocol exchange was observed.
+* Claim values that are not issued by the Issuer cannot be successfully verified by the Verifier,
+  - if the Holder forged the Disclosures.
+
 ## Mandatory digest computation of the revealed claim values by the Verifier
 
 ToDo: add text explaining mechanisms that should be adopted to ensure that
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -599,6 +599,14 @@ Here, `eyJhbGci...emhlaUJhZzBZ` represents the SD-JWT and `eyJhb...dYALCGg` repr
 
 # Security Considerations {#security_considerations}
 
+Security considerations in this section help achieve the following the following properties:
```
bc-pi, line 602:
```suggestion
Security considerations in this section help achieve the following properties:
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -599,6 +599,14 @@ Here, `eyJhbGci...emhlaUJhZzBZ` represents the SD-JWT and `eyJhb...dYALCGg` repr
...
+
+* Claim values issued by the Issuer cannot be obtained by the adversary,
+	- if the Holder did not intend to disclose them,
+	- even if the content of the protocol exchange was observed.
```
bc-pi, line 606: There's not much, if any really, discussion of the second bullet about the exchange being observed in the Security Considerations Section that follows. So I'm not sure it makes sense to call out in this text, which is kinda introducing the Security Considerations.

Sakurann: it's [Confidentiality during Transport](https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#name-confidentiality-during-tran) in privacy considerations that addresses this second bullet point.

it is tricky because encryption is both security and privacy measure.

bc-pi: Yeah, it is tricky. But I don't know about moving that Confidentiality during Transport in into security considerations. I think I'd prefer to just remove that "even if the content of the protocol exchange was observed" part. To keep this new into to the Security Considerations brief and concise.

Sakurann: I have read some other RFCs in oauth WG and the text related to encryption usually tends to be under the privacy considerations section. for example, https://datatracker.ietf.org/doc/html/rfc9068#section-6

so I am ok deleting this bullet point. will reformat because now it does not make sense having bullet point.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -599,6 +599,14 @@ Here, `eyJhbGci...emhlaUJhZzBZ` represents the SD-JWT and `eyJhb...dYALCGg` repr
 
 # Security Considerations {#security_considerations}
 
+Security considerations in this section help achieve the following properties:
+
+* Claim values issued by the Issuer cannot be obtained by the adversary,
+	- if the Holder did not intend to disclose them,
+	- even if the content of the protocol exchange was observed.
+* Claim values that are not issued by the Issuer cannot be successfully verified by the Verifier,
+  - if the Holder forged the Disclosures.
```
Sakurann, line 608:
```suggestion
* Claim values issued by the Issuer cannot be obtained by the adversary, if the Holder did not intend to disclose them.
* Claim values that are not issued by the Issuer cannot be successfully verified by the Verifier, if the Holder forged the Disclosures.
```

bc-pi (approved):

danielfett (approved):

PR #230, opened 2023-03-07
clarify requirements for the combined formats

Sakurann:
It struck me odd that we use an example as a normative definition of the combined formats.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -428,12 +428,14 @@ calculation and the salts. To this end, the Issuer sends the Disclosure objects
 that were also used for the hash calculation, as described in (#creating_disclosures),
 to the Holder.
 
-The data format for sending the SD-JWT and the Disclosures to the Holder is
-as follows:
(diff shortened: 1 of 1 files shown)


bc-pi (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -452,12 +454,14 @@ For Example 1, the Combined Format for Issuance looks as follows:
...
 
-The data format for sending the SD-JWT and the Disclosures to the Verifier is
-as follows (line break added for readability):
+The data format for sending the SD-JWT and the Disclosures to the Verifier MUST be a series of base64url-encoded values, each separated from the next by a single tilde ('~') character as follows:
 
 ```
 <SD-JWT>~<Disclosure 1>~<Disclosure 2>~...~<Disclosure M>~<optional Holder Binding JWT>
 ```
+
+The order of the base64url-encoded values MUST be an SD-JWT, Disclosures, and an optional Holder Binding JWT.
```
danielfett, line 463:
```suggestion
The order of the base64url-encoded values MUST be an SD-JWT, Disclosures, and an optional Holder Binding JWT. In case there is no Holder Binding JWT, the last element is an empty string. The last separating tilde character MUST NOT be omitted.
```

Issue #208, opened 2023-02-09
Terminology: "JWT" / JSON-LD Example


danielfett:
Note to myself:

- SD-JWT is a JWT (or JWS Serialization as discussed)
- Content (in/out) needs to be JSON
- Content does not need specific claims, can be JSON-LD
- We should provide an example for JSON-LD

Sakurann:
PR merged

PR #212, opened 2023-02-10
Example JSON-LD

tlodderstedt:
This is an example of a JSON-LD based credential conveyed in a SD-JWT. SD-JWT provides integrity, authenticity, and selective disclosure for the enclosed JSOMN-LD object. The payload (after processing the SD-JWT) is a JSON-LD object that can be fed into a JSON-LD processor. 
The concrete example is a Vaccination Certificate according to https://w3c-ccg.github.io/vaccination-vocab/.

PR #229, opened 2023-03-07
initial IANA media type registration

Sakurann:
if merged resolves Issue #74.

PR #232, opened 2023-03-10
add example for JSON-LD in the main text

Sakurann:
Has to be merged after PR #212 (will not compile until then, too)
Resolves Issue #208.


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1041,20 +1080,21 @@ pass the following result on to the application for further processing:
 
 <{{examples/complex_ekyc/verified_contents.json}}
 
-## Example 4 - W3C Verifiable Credentials Data Model (work in progress)
+## Example 4a - W3C Verifiable Credentials Data Model v2.0, not using JSON-LD
```
tlodderstedt, line 1083: I suggest to state "using JSON payload" or "JSON claim set" that reads much more positive than "not ..."

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1041,20 +1080,21 @@ pass the following result on to the application for further processing:
...
 
 This example illustrates how the artifacts defined in this specification can be
-represented using W3C Verifiable Credentials Data Model as defined in
-[@VC_DATA].
+represented as W3C Verifiable Credentials Data Model v2.0 [@VC_DATA_v2.0] without using JSON-LD.
```
tlodderstedt, line 1086: again "with the credential content being a set of JSON claims" "with JSON based content".

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1041,20 +1080,21 @@ pass the following result on to the application for further processing:
...
+It uses a content type `credential-claims-set+json` defined in [@VC_JWT], Section 3.
```
tlodderstedt, line 1087: I think the text should also define the `typ` value. Suggest `vc+sd-jwt`. I also suggest to explicitly mention the header name `cty` for the content type.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1041,20 +1080,21 @@ pass the following result on to the application for further processing:
...
 
-SD-JWT is equivalent to an Issuer-signed W3C Verifiable Credential (W3C VC). Disclosures are sent alongside a VC.
+SD-JWT is equivalent to an Issuer-signed W3C Verifiable Credential (W3C VC). Disclosures are sent alongside a W3C VC.
 
-A Presentation with a Holder Binding JWT is equivalent to a Holder-signed W3C Verifiable Presentation (W3C VP).
+A Holder-signed Verifiable Presentation as defined in [@VC_DATA_v2.0] would be equivalent to
```
tlodderstedt, line 1091: "would be" or "is"?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1068,6 +1108,44 @@ Disclosures:
 
 {{examples/w3c-vc/disclosures.md}}
 
+## Example 4b - W3C Verifiable Credentials Data Model v1.1, using JSON-LD
+
+This example illustrates how the artifacts defined in this specification can be
```
tlodderstedt, line 1113: "How the artifacts defined in this presentation can be used to represent a Verifiable Credential according to W3C Verifiable Credentials Data Model v2.0 ..."

tlodderstedt (changes requested):
I miss the definition of the `typ` value for SD-JWT-based VCs. Suggest `vc+sd-jwt`. I also made some suggestions for text modifications.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: `would be` because we are not putting the example of a VP.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: paraphrased

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: JSON-LD is JSON, too. We say `without using JSON-LD` in OID4VP spec. I want to be explicit.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: JSON-LD is JSON, too. We say `without using JSON-LD` in OID4VP spec. I want to be explicit that it is not JSON-LD

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: yes, `typ` should be defined for sd-jwt-vc, I added for clarity, but I strongly believe then it should be in the main text and not the examples that are **non-normative**.

Sakurann: `cty` clarified.

bc-pi: Non-normative examples should not be using (or pseudo defining) nonexistent media types.  

`vc+sd-jwt` isn't a registered media type nor is there text anywhere that's intending to define/register it. It also suggests a structured-suffix media type of `+sd-jwt` which similarly doesn't exist. Those maybe things that'd be desirable to define/register but not just by putting them in examples. That's a somewhat different direction than what's being proposed in #229 too.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -931,10 +955,25 @@ TBD
...
+    <author fullname="Kyle Den Hartog">
+      <organization>MATTR</organization>
+    </author>
+    <date day="03" month="Mar" year="2022" />
+  </front>
+</reference>
+
+@VC_JWT
+
+<reference anchor="VC_JWT" target="https://w3c.github.io/vc-jwt/">
```
OR13, line 967: Beware that this will change to https://www.w3.org/TR/vc-jwt when it goes to FPWD.

You might want to file an issue to track fixing it if you fear you will forget.

draft-ietf-oauth-selective-disclosure-jwt.md
OR13: I agree with @tlodderstedt ... The example and text should be what makes it explicit.

perhaps "W3C Verifiable Credentials as Native JSON Web Tokens" ?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1041,20 +1080,21 @@ pass the following result on to the application for further processing:
...
 
-## Example 4 - W3C Verifiable Credentials Data Model (work in progress)
+## Example 4a - W3C Verifiable Credentials Data Model v2.0, not using JSON-LD
 
-This example illustrates how the artifacts defined in this specification can be
-represented using W3C Verifiable Credentials Data Model as defined in
-[@VC_DATA].
+This example illustrates how to use the artifacts defined in this specification to secure a payload
+that is represented as a W3C Verifiable Credentials Data Model v2.0 [@VC_DATA_v2.0]
+and does not use JSON-LD. The example uses a content type `credential-claims-set+json` defined in [@VC_JWT], Section 3 in a `cty` JOSE Header value.
```
OR13, line 1087:
```suggestion
and does not use JSON-LD to provide semantic definitions for claims. The example uses a content type `credential-claims-set+json` defined in [@VC_JWT], Section 3 in a `cty` JOSE Header value.
```

draft-ietf-oauth-selective-disclosure-jwt.md
OR13: will be with an issue marker ?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1041,20 +1080,21 @@ pass the following result on to the application for further processing:
...
-SD-JWT is equivalent to an Issuer-signed W3C Verifiable Credential (W3C VC). Disclosures are sent alongside a VC.
+SD-JWT is equivalent to an Issuer-signed W3C Verifiable Credential (W3C VC). Disclosures are sent alongside a W3C VC.
 
-A Presentation with a Holder Binding JWT is equivalent to a Holder-signed W3C Verifiable Presentation (W3C VP).
+A Holder-signed Verifiable Presentation as defined in [@VC_DATA_v2.0] would be equivalent to
+a Combined Format for Presentation with a Holder Binding JWT.
 
-Holder Binding is applied and the Holder Binding JWT is signed using a raw public key passed in a `cnf` Claim in a W3C VC (SD-JWT).
+In this example, Holder Binding is applied and Verifiable Presentation can be signed using
+a raw public key passed in a `cnf` Claim in the SD-JWT.
```
OR13, line 1095: not sure we want to call JWK a "raw public key".... I think `cnf` also supports by reference / URL... do we want to support both?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1068,6 +1108,44 @@ Disclosures:
 
 {{examples/w3c-vc/disclosures.md}}
 
+## Example 4b - W3C Verifiable Credentials Data Model v2.0, using JSON-LD
+
+This example illustrates how to use the artifacts defined in this specification to secure a payload
+that is represented as a W3C Verifiable Credentials Data Model v2.0 [@VC_DATA_v2.0]
+and uses JSON-LD. The example uses a content type `credential+ld+json` defined in [@VC_DATA_v2.0], Section 6.3 in a `cty` JOSE Header value.
```
OR13, line 1115:
```suggestion
and uses a JSON-LD object as the the claims set. The example uses a content type `credential+ld+json` defined in [@VC_DATA_v2.0], Section 6.3 in a `cty` JOSE Header value.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1068,6 +1108,44 @@ Disclosures:
...
+
+SD-JWT is equivalent to an Issuer-signed W3C Verifiable Credential (W3C VC). Disclosures are sent alongside a VC.
+
+A Combined Format for Presentation with a Holder Binding JWT would be equivalent to a Holder-signed
+Verifiable Presentation as defined in [@VC_DATA_v2.0].
+
+In this example, Holder Binding is applied and a Combined Format for Presentation is signed
+using a raw public key passed in a `cnf` Claim in the SD-JWT.
```
OR13, line 1123: same comments, if these sections going to be identical, it would be nice to be DRY-er.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1068,6 +1108,44 @@ Disclosures:
...
+
+<{{examples/jsonld/sd_jwt_payload.json}}
+
+Disclosures:
+
+{{examples/jsonld/disclosures.md}}
+
+A Combined Format for Presentation for the SD-JWT that discloses only discloses `type`, `medicinalProductName`, `atcCode` of the vaccine,
+`type` of the `recipient`, `type`, `order` and `dateOfVaccination`,
+and has a Holder Binding JWT could look as follows:
```
OR13, line 1141: This reads a bit awkward, I think it would be nice to split the claims disclosure from the confirmation binding, but leave both as traits of the example.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -55,10 +55,10 @@ secured against modification using JSON Web Signature (JWS) [@!RFC7515] digital
 signatures. A consumer of a signed JWT that has checked the
 signature can safely assume that the contents of the token have not been
 modified.  However, anyone receiving an unencrypted JWT can read all the
-claims and likewise, anyone with the decryption key receiving an encrypted JWT
+claims. Likewise, anyone with the decryption key receiving a signed and encrypted JWT
 can also read all the claims.
 
-One of the common use cases of a signed JWT is representing a user's
+One of the common use cases of a signed JWT is representing user's
```
OR13, line 61:
```suggestion
One of the common use cases of a signed JWT is representing a user's
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -102,6 +102,8 @@ or the concept of binding an SD-JWT to key material controlled by the
 Holder. The strength of the Holder Binding is conditional upon the trust
 in the protection of the private key of the key pair an SD-JWT is bound to.
 
+SD-JWT can be used with any JSON-based representation of claims, including JSON-LD.
```
OR13, line 105: I like this.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1041,20 +1080,21 @@ pass the following result on to the application for further processing:
...
 
-This example illustrates how the artifacts defined in this specification can be
-represented using W3C Verifiable Credentials Data Model as defined in
-[@VC_DATA].
+This example illustrates how to use the artifacts defined in this specification to secure a payload
+that is represented as a W3C Verifiable Credentials Data Model v2.0 [@VC_DATA_v2.0]
+and does not use JSON-LD. The example uses a content type `credential-claims-set+json` defined in [@VC_JWT], Section 3 in a `cty` JOSE Header value.
 
-SD-JWT is equivalent to an Issuer-signed W3C Verifiable Credential (W3C VC). Disclosures are sent alongside a VC.
+SD-JWT is equivalent to an Issuer-signed W3C Verifiable Credential (W3C VC). Disclosures are sent alongside a W3C VC.
```
OR13, line 1089: the equivalence language reads roughly...  i think there are 2 cases:

SD-JWT is a VC where:

1. the verified / decoded claim set is a core data model media type. (with JSON-LD)
2. the verified / decoded claim set JSON. (without JSON-LD)

Both are "W3C VCs" if a mapping exists for 2.

It would be preferable to call them both "W3C Verifiable Credentials", and define a minimal mapping to ensure 2 is valid, with caveats, like we are doing here: https://github.com/w3c/vc-jwt/pull/60

OR13 (approved):
All my suggestions are nits.

I support merging the PR with or without addressing them, it improves the document.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -55,10 +55,10 @@ secured against modification using JSON Web Signature (JWS) [@!RFC7515] digital
 signatures. A consumer of a signed JWT that has checked the
 signature can safely assume that the contents of the token have not been
 modified.  However, anyone receiving an unencrypted JWT can read all the
-claims and likewise, anyone with the decryption key receiving an encrypted JWT
+claims. Likewise, anyone with the decryption key receiving a signed and encrypted JWT
```
danielfett, line 58: The 'signed' part is besides the point (we did say only 'unencrypted' in the previous sentence) because that's what matters here.

danielfett (changes requested):

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
a Holder's public key passed in a `cnf` Claim in the SD-JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
using a Holder's public key passed in a `cnf` Claim in the SD-JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
A Holder-signed Verifiable Presentation as defined in [@VC_DATA_v2.0] is equivalent to
```

danielfett (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
claims. Likewise, anyone with the decryption key receiving encrypted JWT
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: thanks, issue #237

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I will merge the current language as a starting point. can you please file an issue on this so that we can fix it later?

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann, line 1141:
```suggestion
The following is a non-normative example of a Combined Format for Presentation for the SD-JWT with a Holder Binding JWT. It only discloses `type`, `medicinalProductName`, `atcCode` of the vaccine, `type` of the `recipient`, `type`, `order` and `dateOfVaccination`.
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: keeping the language for now after discussion with other editors. please file an issue if you feel strongly about it.

bc-pi (approved):

PR #249, opened 2023-04-11
initial IANA media type and structured suffix registration

bc-pi:


Sakurann:
Will close #229 when this PR gets merged

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -914,6 +914,51 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
 
 TBD
 
+## Media Type Registration
+
+This section requests registration of the "application/sd-jwt" media type [@RFC2046] in
```
Sakurann, line 919: @OR13 @tlodderstedt when we discussed in person, how did we say this one media type will be used for both issuance and presentation of SD-JWT? Those will have different processing rules (section 6) and the latter might have HB JWT too...

Sakurann (commented):
Direction lgtm. I am just getting confused about application/sd-jwt (again)

draft-ietf-oauth-selective-disclosure-jwt.md
OR13: I don't recall, but if it is in the protected header, and must be disclosed, it can't be changed after issuance.

There won't be any integrity protection on a "presentation format token" unless it is signed over by the holder, so as you say, holder binding JWT probably wants a specific sub type... something like:

`application/hb-sd-jwt`

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -914,6 +914,51 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
...
+the "Media Types" registry [@IANA.MediaTypes] in the manner described
+in [@RFC6838], which can be used to indicate that the content is an SD-JWT.
+
+* Type name: application
+* Subtype name: sd-jwt
+* Required parameters: n/a
+* Optional parameters: n/a
+* Encoding considerations: binary; application/sd-jwt values are a series of base64url-encoded values (some of which may be the empty string) separated by period ('.') or tilde ('~') characters.
+* Security considerations: See the Security Considerations section of [[ this specification ]].
```
OR13, line 928: imo, you should also reference security considerations of RFC7519 here.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -914,6 +914,51 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
...
+the manner described in [RFC6838], which can be used to indicate that
+the media type is encoded as an SD-JWT.
+
+* Name: SD-JWT
+* +suffix: +sd-jwt
+* References: [[ this specification ]]
+* Encoding considerations: binary; SD-JWT values are a series of base64url-encoded values (some of which may be the empty string) separated by period ('.') or tilde ('~') characters.
+* Interoperability considerations: n/a
+* Fragment identifier considerations: n/a
+* Security considerations: See the Security Considerations section of [[ this specification ]].
```
OR13, line 957: Same comment about referencing RFC7519 as well, or maybe just note that at the top of that section...

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: Makes sense. I'll add that. And maybe the JWT BCP RFC8725 too.

bc-pi: done

bc-pi:
Does there need to be anything said for these in the main body of the document?

draft-ietf-oauth-selective-disclosure-jwt.md
tlodderstedt: The media type is always the same (as it is signed by the issuer). The additional hb JWT does not seem to have a dedicated `typ` yet, but I think it should have one. `application/hb+jwt` is reasonable to me. 

What do the co-authors think? @bc-pi @Sakurann @danielfett

tlodderstedt (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: `application/hb+jwt` is reasonable to me too but I'm not sure the HB JWT needs to be typed.

OR13: @bc-pi yea, it does not need to be typed, unless you want to ensure header or claimset members are processed a specific way... the `nonce` seems like it might be one of those... not sure.

OR13 (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: ok, so do I understand that the same application/sd-jwt will be used regardless whether it is issuance and presentation and the receiver is expected to know which combined format it is, based on the context, protocol?

bc-pi: Yes @Sakurann, I believe so.

Sakurann (approved):

danielfett (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1281,6 +1339,10 @@ data. The original JSON data is then used by the application. See
 
    [[ To be removed from the final specification ]]
 
+   -05
+
+   * Added initial IANA media type and structured suffix registration requests
```
danielfett, line 1344:
```suggestion
   * Added initial IANA media type and structured suffix registration requests
   * Added considerations around forwarding credentials
```

Issue #235, opened 2023-03-13
Add more details on attacker model


danielfett:
Issue #87 discusses the security and privacy targets. This was fixed in PR #225, but we should add more details on the attacker model.

bc-pi:
PR was merged

PR #250, opened 2023-04-14
Attempt to improve description of security properties

danielfett:
This fixes Issue #235. The main  change is to better describe in which role the adversary acts.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -621,8 +621,16 @@ Other specifications or profiles of this specification may define alternative fo
 
 Security considerations in this section help achieve the following properties:
 
-* Claim values issued by the Issuer cannot be obtained by the adversary, if the Holder did not intend to disclose them.
-* Claim values that are not issued by the Issuer cannot be successfully verified by the Verifier, if the Holder forged the Disclosures.
+**Selective Disclosure:** An adversary in the role of the Verifier cannot obtain
+information from an SD-JWT about any claim name or claim value that was not
+explicitly disclosed by the Holder unless that information can be derived from
+other disclosed claims or external sources.
+
+**Integrity:** The claims verified by a Verifier are a subset of the claims
+released by the Issuer, even if the Holder is malicious (i.e., a Holder cannot
+modify claim names or claim values without the Verifier noticing during verification).
+
+Additionally, Holder Binding can be achieved as described in (#holder_binding_security).
 
 ## Mandatory signing of the SD-JWT
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -624,7 +624,7 @@ Security considerations in this section help achieve the following properties:
 **Selective Disclosure:** An adversary in the role of the Verifier cannot obtain
 information from an SD-JWT about any claim name or claim value that was not
 explicitly disclosed by the Holder unless that information can be derived from
-other disclosed claims or external sources.
+other disclosed claims or sources other than the presented SD-JWT.
 
 **Integrity:** A malicious Holder cannot modify names or values of selectively disclosable claims without detection by the Verifier.
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -628,7 +628,7 @@ other disclosed claims or external sources.
 
 **Integrity:** A malicious Holder cannot modify names or values of selectively disclosable claims without detection by the Verifier.
 
-Additionally, Holder Binding can be achieved as described in (#holder_binding_security).
+Additionally, as described in (#holder_binding_security), the application of Holder Binding can ensure that the presenter of an SD-JWT credential is the legitimate Holder of the credential.
 
 ## Mandatory signing of the SD-JWT
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -626,9 +626,7 @@ information from an SD-JWT about any claim name or claim value that was not
 explicitly disclosed by the Holder unless that information can be derived from
 other disclosed claims or external sources.
 
-**Integrity:** The claims verified by a Verifier are a subset of the claims
-released by the Issuer, even if the Holder is malicious (i.e., a Holder cannot
-modify claim names or claim values without the Verifier noticing during verification).
+**Integrity:** A malicious Holder cannot modify names or values of selectively disclosable claims without detection by the Verifier.
 
 Additionally, Holder Binding can be achieved as described in (#holder_binding_security).
 


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -621,8 +621,16 @@ Other specifications or profiles of this specification may define alternative fo
...
-* Claim values issued by the Issuer cannot be obtained by the adversary, if the Holder did not intend to disclose them.
-* Claim values that are not issued by the Issuer cannot be successfully verified by the Verifier, if the Holder forged the Disclosures.
+**Selective Disclosure:** An adversary in the role of the Verifier cannot obtain
+information from an SD-JWT about any claim name or claim value that was not
+explicitly disclosed by the Holder unless that information can be derived from
+other disclosed claims or external sources.
+
+**Integrity:** The claims verified by a Verifier are a subset of the claims
+released by the Issuer, even if the Holder is malicious (i.e., a Holder cannot
+modify claim names or claim values without the Verifier noticing during verification).
```
bc-pi, line 631: I'm finding the main sentence kinda hard to process. What about just something inspired by the text you had in the parenthetical? 

```suggestion
**Integrity:** A malicious Holder cannot modify names or values of selectively disclosable claims without detection by the Verifier.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -621,8 +621,16 @@ Other specifications or profiles of this specification may define alternative fo
 
 Security considerations in this section help achieve the following properties:
 
-* Claim values issued by the Issuer cannot be obtained by the adversary, if the Holder did not intend to disclose them.
-* Claim values that are not issued by the Issuer cannot be successfully verified by the Verifier, if the Holder forged the Disclosures.
+**Selective Disclosure:** An adversary in the role of the Verifier cannot obtain
```
Sakurann, line 624:
```suggestion
* An adversary in the role of the Verifier cannot obtain
```
I find two abstractions of a property achieved by security considerations confusing.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I like Brian's suggested text. Would also suggest removing **Integrity**. Per a comment above, I find two introduced abstractions confusing.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -621,8 +621,16 @@ Other specifications or profiles of this specification may define alternative fo
...
+information from an SD-JWT about any claim name or claim value that was not
+explicitly disclosed by the Holder unless that information can be derived from
+other disclosed claims or external sources.
+
+**Integrity:** The claims verified by a Verifier are a subset of the claims
+released by the Issuer, even if the Holder is malicious (i.e., a Holder cannot
+modify claim names or claim values without the Verifier noticing during verification).
+
+Additionally, Holder Binding can be achieved as described in (#holder_binding_security).
```
Sakurann, line 633: Suggest removing this text. Holder binding ensures that it is the holder that directly received combined format for issuance that is taking to the verifier. But as Orie's PR suggests, some use-cases might want to allow it.

```suggestion
```

Sakurann (changes requested):

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: Not sure what you mean by abstractions, but it is usually helpful to introduce names for the security properties you want to achieve. I'm open to suggestions for better names.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: I mention holder binding here as something you might want to achieve, as it is another important security property.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: is "selective disclosure" a security property..?

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: then it should be explained more, including a security property HB achieves

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: It is if we define it, and I think we should because that's the main goal of the spec. I'm open to better names, though.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: Maybe something like this then? Tries to explain it a bit while saying what it achieves.  

```suggestion
Additionally, as described in (#holder_binding_security), the application of Holder Binding can ensure that the presenter of an SD-JWT credential is the legitimate Holder of the credential.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -621,8 +621,14 @@ Other specifications or profiles of this specification may define alternative fo
 
 Security considerations in this section help achieve the following properties:
 
-* Claim values issued by the Issuer cannot be obtained by the adversary, if the Holder did not intend to disclose them.
-* Claim values that are not issued by the Issuer cannot be successfully verified by the Verifier, if the Holder forged the Disclosures.
+**Selective Disclosure:** An adversary in the role of the Verifier cannot obtain
+information from an SD-JWT about any claim name or claim value that was not
+explicitly disclosed by the Holder unless that information can be derived from
+other disclosed claims or external sources.
```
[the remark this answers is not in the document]
danielfett: For example, if `nationalities` is selectively disclosable, and the holder does not disclose it to the verifier, the verifier might still learn the nationality from 

 - a passport image acquired out of band, e.g., during a video identification session (external source), or
 - a disclosed claim containing the passport number, which may leak information about the nationality ("other disclosed claims").

So we have to say "the verifier does not learn the `nationality` claim's value from the presentation of that claim" but it still may learn it from somewhere else because we cannot control that.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -621,8 +621,14 @@ Other specifications or profiles of this specification may define alternative fo
...
+
+**Integrity:** A malicious Holder cannot modify names or values of selectively disclosable claims without detection by the Verifier.
+
+Additionally, Holder Binding can be achieved as described in (#holder_binding_security).
```
danielfett, line 631: Maybe we should use the same format as for the other two properties. Re-using text from Brian's proposal:

```suggestion
**Holder Binding:** When Holder Binding is enforced, only the legitimate Holder of the credential can present the SD-JWT (as described in (#holder_binding_security)).
```

Sakurann: I would be ok with 
> **Holder Binding:** Only the legitimate Holder of the credential can present the SD-JWT (as described in (#holder_binding_security)).

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann, line 627:
```suggestion
explicitly disclosed by the Holder unless that information can be derived from
other disclosed claims or is available from sources other than SD-JWT.
```

danielfett: How about this?
```suggestion
explicitly disclosed by the Holder unless that information can be derived from
other disclosed claims or sources other than the presented SD-JWT.
```

Sakurann (approved):

bc-pi (approved):

Issue #240, opened 2023-03-27
Example 2 mismatch


christianpaquin:
[Example 2b](https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#name-example-2b-adding-decoys) states that it was created using data from [Example 2a](https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#name-example-2a-handling-structu), but adding decoy values. I'd expect its `_sd` param to be a superset of example 2a's, but it's not the case. Am I missing something?

danielfett:
Example 2b includes different nonces than Example 2a.

We try to use deterministic random numbers to generate the examples, but in this case, due to the order in which the decoys and regular salts are created, this approach does not produce the same nonces for the same "real" disclosures.

bc-pi:
PR was merged

PR #251, opened 2023-04-14
Remove Example 2b to fix Issue #240

danielfett:
To fix Issue #240, I found that removing Example 2b is the best way to go. 

I tried to fixate the nonces used in the examples despite decoys being generated in one of the examples, but that got quite messy and introduces a lot of unsafe code paths into the python code.

An alternative would be to manage the example manually, but that is a source for errors, so let's not do that.


bc-pi:
The decoy concept is intuitive enough that a separate example isn't needed, I think. And per #240 the separate example that has a completely different set of digests might cause confusion. So removing 2b seems like a good course of action. 

The change is larger than I'd have expected. Are all the other file changes due to regenerating examples? I think so but it caught my attention. And it seems like #252 and this will be difficult to resolve. But maybe you've already considered that.

Sakurann:
Why remove an entire example, if we can add a text clarifying that "Example 2b uses the same data as 2a, but because even for the same claim values, different nonces are used, not just the decoys but all digests are different from those in 2a"?

danielfett:
That was my initial approach, but with that change there's really not that much value to the example: It just shows a bunch of digests, where the number of digests is higher than it was previously. It can be confusing for readers at worst and is not very helpful at best.

Sakurann (changes requested):
given feedback in issue #260. I am not still not sure entirely removing is a good option...
Could we add to example 2 a text saying "if the issuer wants to use decoys, these strings would be added to an `_sd` array: <2-3 decoy digests>"?

danielfett:
So instead of showing the decoy digests in the examples you propose to list them seperately and not show a complete example with decoy digests?

danielfett:
We agreed on our call to generate a snippet of markdown code to explicitly list the decoy digests to improve the explanation.

danielfett:
I updated the python script and the spec to show the decoy digests: https://drafts.oauth.net/oauth-selective-disclosure-jwt/danielfett/example-2b-description/draft-ietf-oauth-selective-disclosure-jwt.html#name-example-2-handling-structur

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1123,37 +1123,29 @@ the media type is encoded as an SD-JWT.
...
 
-## Example 2a: Handling Structured Claims {#example-simple_structured}
+## Example 2: Handling Structured Claims {#example-simple_structured}
 
 This example uses the following object as the set of claims that the Issuer is issuing:
 
 <{{examples/simple_structured/user_claims.json}}
 
-Note that in contrast to Example 1, here the Issuer decided to create a structured object for the `address` claim, allowing for separate disclosure of the individual members of the claim.
+In contrast to Example 1, here the Issuer decided to create a structured object for the `address` claim, allowing for separate disclosure of the individual members of the claim, and also added decoy digests to prevent the Verifier from deducing the true number of claims in the address object. The following payload is used for the SD-JWT:
```
bc-pi, line 1132: The example has decoys in the address and at the top level too. So this text should probably be more general about the decoys.  Maybe just:
```suggestion
In contrast to Example 1, here the Issuer decided to create a structured object for the `address` claim, allowing for separate disclosure of the individual members of the claim, and also added decoy digests to prevent the Verifier from deducing the true number of claims. The following payload is used for the SD-JWT:
```

bc-pi:
Other than the little nit about the the text that says the decoys in the example are about the address, this looks pretty good.

bc-pi (approved):

bc-pi:
Do you want to mention something in the doc history?

danielfett:
Good point, done!

danielfett:
@Sakurann Can you please review this?

Sakurann (approved):

PR #262, opened 2023-05-08
define HB-JWT normatively

Sakurann:
Discussions around defining SD-JWT-VC show that it would be very helpful for interoperability if HB JWT is normatively defined in SD-JWT specification itself. Using HB JWT is optional, but if used, it should be compliant to the definitions in SD-JWT.

How to represent validity (iat, exp, nbf) will probably need the most discussion...

also we should introduce an abbreviation `HB JWT`....

cc @awoie


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -485,16 +485,29 @@ a Disclosure more than once.
...
+
+This section defines the contents of the Holder Binding JWT.
+
+The JWT MUST contain the following elements:
+  * in the JOSE header,
+    * `typ`: REQUIRED. MUST be `hb+jwt`, which explicitly types the Holder Binding JWT as recommended in Section 3.11 of [@!RFC8725].
+    * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+    * `kid`: `kid`: CONDITIONAL. JOSE Header containing the key ID. MUST NOT be present if `jwk` or `x5c` is present.
+    * `jwk`: CONDITIONAL. JOSE Header containing the key material the new Credential shall be bound to. MUST NOT be present if `kid` or `x5c` is present.
+    * `x5c`: CONDITIONAL. JOSE Header containing a certificate or certificate chain corresponding to the key used to sign the JWT. This element MAY be used to convey a key attestation. In such a case, the actual key certificate will contain attributes related to the key properties. MUST NOT be present if `kid` or `jwk` is present.
```
bc-pi, line 499: I'm not sure I really grok the key attestation conveyance stuff, to be honest.

bc-pi, line 499: In general I think the JOSE headers that convey or point to the key don't need to be and shouldn't be in this list. The HB-JWT for an SD-JWT that is holder bound with a cnf.jwk doesn't need any of these, for one example. And the requirements re kid/jwk/x5c as written in these bullets now aren't totally clear (to me anyway) and need some fixup. Or just remove them.  I'd argue that the defined structure of the Holder Binding JWT should be only the minimal claims/headers necessary. Other requirements, if needed, such as saying that no more than one of `kid` or `jwk` or `x5c` be used together could be in text elsewhere.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -485,16 +485,29 @@ a Disclosure more than once.
...
-Example Holder Binding JWT payload:
+#### Structure of the Holder Binding JWT {#hb-jwt}
+
+This section defines the contents of the Holder Binding JWT.
+
+The JWT MUST contain the following elements:
+  * in the JOSE header,
+    * `typ`: REQUIRED. MUST be `hb+jwt`, which explicitly types the Holder Binding JWT as recommended in Section 3.11 of [@!RFC8725].
+    * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+    * `kid`: `kid`: CONDITIONAL. JOSE Header containing the key ID. MUST NOT be present if `jwk` or `x5c` is present.
```
bc-pi, line 497: "`kid`: `kid`: "  should probably only have one "`kid`:"

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -485,16 +485,29 @@ a Disclosure more than once.
...
+    * `jwk`: CONDITIONAL. JOSE Header containing the key material the new Credential shall be bound to. MUST NOT be present if `kid` or `x5c` is present.
+    * `x5c`: CONDITIONAL. JOSE Header containing a certificate or certificate chain corresponding to the key used to sign the JWT. This element MAY be used to convey a key attestation. In such a case, the actual key certificate will contain attributes related to the key properties. MUST NOT be present if `kid` or `jwk` is present.
+  * in the JWT body,
+    * `iss`: REQUIRED. The issuer of the Holder Binding JWT. How the value is represented is up to the protocol used and out of scope of this specification.
```
bc-pi, line 501: I don't think issuer should be required. It's kinda implied by context to be the Holder. And doesn't really add anything to have it, that I can see. And there may not always be an obvious value to use for the iss value for the holder. Seems more likely to hurt than help.

bc-pi: Oh and also the current examples of the Holder Binding JWT or its payload have iat, aud, and nonce. But not iss. Removing it here, where it's not necessary IMHO, would be easier than updating the code and examples. 

Also I suspect there's some reason Dr. Fett didn't have `iss` in the HB JWT payloads. Even if not normative examples.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -485,16 +485,29 @@ a Disclosure more than once.
...
-other claims are obtained by the Holder is out of scope of this specification.
 
-Example Holder Binding JWT payload:
+#### Structure of the Holder Binding JWT {#hb-jwt}
+
+This section defines the contents of the Holder Binding JWT.
+
+The JWT MUST contain the following elements:
+  * in the JOSE header,
+    * `typ`: REQUIRED. MUST be `hb+jwt`, which explicitly types the Holder Binding JWT as recommended in Section 3.11 of [@!RFC8725].
```
bc-pi, line 495: So... the encoded HB-JWT examples (or the code that generates them) should probably be updated to have this `typ`

bc-pi (changes requested):
This change probably also warrants a mention in the doc history - i.e. just a simple, "Defined the structure of the Holder Binding JWT"

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: > Also I suspect there's some reason Dr. Fett didn't have `iss` in the HB JWT payloads. Even if not normative examples.

Essentially what you said above - (1) I'm not sure if there's always a defined value to put there and (2) I don't think it is necessary. Audience is really important, but I don't see a strict need for iss.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -485,16 +485,29 @@ a Disclosure more than once.
...
+    * `iat`: REQUIRED. The value of this claim MUST be the time at which the Holder Binding JWT was issued using the syntax defined in [@!RFC7519].
+    * `aud`: REQUIRED. The intended receiver of the Holder Binding JWT. How the value is represented is up to the protocol used and out of scope of this specification.
+    * `nonce`: REQUIRED. Ensures the freshness of the signature. The value type of this claim MUST be a string. How this value is obtained is up to the protocol used and out of scope of this specification.
+
+Below is a non-normative example of a Holder Binding JWT payload:
 
 <{{examples/simple/hb_jwt_payload.json}}
 
-Which is then signed by the Holder to create a JWT like the following:
+Below is a non-normative example of a JWT produced by signign a payload in the example above:
```
danielfett, line 510:
```suggestion
Below is a non-normative example of a JWT produced by signing a payload in the example above:
```

danielfett (changes requested):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -485,16 +485,29 @@ a Disclosure more than once.
...
+#### Structure of the Holder Binding JWT {#hb-jwt}
+
+This section defines the contents of the Holder Binding JWT.
+
+The JWT MUST contain the following elements:
+  * in the JOSE header,
+    * `typ`: REQUIRED. MUST be `hb+jwt`, which explicitly types the Holder Binding JWT as recommended in Section 3.11 of [@!RFC8725].
+    * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+    * `kid`: `kid`: CONDITIONAL. JOSE Header containing the key ID. MUST NOT be present if `jwk` or `x5c` is present.
+    * `jwk`: CONDITIONAL. JOSE Header containing the key material the new Credential shall be bound to. MUST NOT be present if `kid` or `x5c` is present.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
    * `jwk`: CONDITIONAL. JOSE Header containing the key material. MUST NOT be present if `kid` or `x5c` is present.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -582,8 +595,8 @@ To this end, Verifiers MUST follow the following steps (or equivalent):
        2. Determine the public key for the Holder from the SD-JWT.
        3. Ensure that a signing algorithm was used that was deemed secure for the application. Refer to [@RFC8725], Sections 3.1 and 3.2 for details. The `none` algorithm MUST NOT be accepted.
        4. Validate the signature over the Holder Binding JWT.
-       5. Check that the Holder Binding JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the Holder Binding JWT.
-       6. Determine that the Holder Binding JWT is bound to the current transaction and was created for this Verifier (replay protection). This is usually achieved by a `nonce` and `aud` field within the Holder Binding JWT.
+       5. Check that the Holder Binding JWT is valid using `iat`, and, if provided, `nbf` and `exp` claims.
+       6. Determine that the Holder Binding JWT is bound to the current transaction and was created for this Verifier (replay protection) by validating `nonce` and `aud` claims.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
       5. Check that the Holder Binding JWT is valid using `iat`, and `typ` and, if provided, `nbf` and `exp` claims.
       6. Determine that the Holder Binding JWT is bound to the current transaction and was created for this Verifier (replay protection) by validating `nonce` and `aud` claims.
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
```

draft-ietf-oauth-selective-disclosure-jwt.md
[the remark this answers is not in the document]
Sakurann: ok to remove JOSE headers and replace with a sentence saying "use the key identified in SD-JWT, if not clear which one, applications of sd-jwt must specify which key to use in HB-JWT."

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I updated the code and added examples, but did not re-generate example 4b, and 4a (was too much)....

Sakurann:
need to re-generate example 4b, and 4a to update hb jwt header....

Sakurann:
cc @awoie please take a look (can't ask review from you...)

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -485,16 +485,36 @@ a Disclosure more than once.
...
+  * in the JOSE header,
+    * `typ`: REQUIRED. MUST be `hb+jwt`, which explicitly types the Holder Binding JWT as recommended in Section 3.11 of [@!RFC8725].
+    * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
+  * in the JWT body,
+    * `iat`: REQUIRED. The value of this claim MUST be the time at which the Holder Binding JWT was issued using the syntax defined in [@!RFC7519].
+    * `aud`: REQUIRED. The intended receiver of the Holder Binding JWT. How the value is represented is up to the protocol used and out of scope of this specification.
+    * `nonce`: REQUIRED. Ensures the freshness of the signature. The value type of this claim MUST be a string. How this value is obtained is up to the protocol used and out of scope of this specification.
+
+To validate the signature on the Holder Binding JWT, the Verifier MUST use the key material in the SD-JWT. If it is not clear from the SD-JWT, HB-JWT MUST specify which key material the Verifier needs to use to validate HB-JWT using JOSE header parameters such as `kid` and `x5c`.
```
awoie, line 502: @Sakurann Does that mean I can use a DID as well? I don't want to be too restrictive in what can be used for verifying the HB-JWT? Being restrictive is the job of specific profiles such as the hi-assurance profile.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -582,8 +602,9 @@ To this end, Verifiers MUST follow the following steps (or equivalent):
        2. Determine the public key for the Holder from the SD-JWT.
        3. Ensure that a signing algorithm was used that was deemed secure for the application. Refer to [@RFC8725], Sections 3.1 and 3.2 for details. The `none` algorithm MUST NOT be accepted.
        4. Validate the signature over the Holder Binding JWT.
-       5. Check that the Holder Binding JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the Holder Binding JWT.
-       6. Determine that the Holder Binding JWT is bound to the current transaction and was created for this Verifier (replay protection). This is usually achieved by a `nonce` and `aud` field within the Holder Binding JWT.
+       5. Check that the `typ` of the Holder Binding JWT is `hb+jwt`.
```
awoie, line 605: Why is `hb+jwt` needed for this? We would also need to register a media type -> application/hb+jwt for this.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -582,8 +602,9 @@ To this end, Verifiers MUST follow the following steps (or equivalent):
...
+       6. Check that the Holder Binding JWT is valid using `iat`, and, if provided, `nbf` and `exp` claims.
```
awoie, line 606: If `nbf` and `exp` are not provided, then I think this step is not needed. How would somebody check whether the HB-JWT is valid if only `iat` is provided`

awoie (changes requested):
direction looks great, just a few questions and comments.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: yes, that's the plan, to register

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I think the idea is that the Verifier wants HB-JWTs that are relatively freshly created. And so would reject HB-JWTs that have an `iat` that falls outside some policy based window.

Sakurann: it's the same logic as [DPoP JWT](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop-16#section-4.2). iat is sufficient to know if that HB-JWT is fresh enough.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I don't think anything in the text prevents DIDs. `kid` is expected to be used mainly with DIDs.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I think the idea is that the Verifier wants HB-JWTs that are relatively freshly created. And so would reject HB-JWTs that have an `iat` that falls outside some policy based window.  But something like that should probably said more explicitly because `iat` doesn't have any normative validity rules on its own. Maybe, "Check that Holder Binding JWT was created within an acceptable timeframe as indicated by the `iat` claim" or something along those lines.

FWIW the idea is similar to DPoP and [Checking DPoP Proofs](https://www.ietf.org/archive/id/draft-ietf-oauth-dpop-16.html#name-checking-dpop-proofs) does have some text that attempts to say the the creation time of the JWT needs to be within an acceptable window.

And rather than mentioning the`nbf` and `exp` claims maybe have another general item like "Ensure that the Holder Binding JWT is valid in all other respects per [@!RFC7519]."

draft-ietf-oauth-selective-disclosure-jwt.md
awoie: sounds good.

draft-ietf-oauth-selective-disclosure-jwt.md
awoie: ok, got it

awoie (approved):
lgtm

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
       6. Check that the creation time of the Holder Binding JWT, as determined by the `iat` claim, is within an acceptable window.
       7. Check that the Holder Binding JWT is valid in all other respects, per [@!RFC7519] and [@!RFC8725].
```
updated the verfication steps as suggested.
@awoie @bc-pi

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I added the IANA registration section in this PR :)

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: wfm

bc-pi:
This change is sufficient to warrant an entry in the doc history like just, "Defined the structure of the Holder Binding JWT"

Sakurann:
done, thank you, Brian

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -582,8 +602,10 @@ To this end, Verifiers MUST follow the following steps (or equivalent):
        2. Determine the public key for the Holder from the SD-JWT.
        3. Ensure that a signing algorithm was used that was deemed secure for the application. Refer to [@RFC8725], Sections 3.1 and 3.2 for details. The `none` algorithm MUST NOT be accepted.
        4. Validate the signature over the Holder Binding JWT.
-       5. Check that the Holder Binding JWT is valid using `nbf`, `iat`, and `exp` claims, if provided in the Holder Binding JWT.
-       6. Determine that the Holder Binding JWT is bound to the current transaction and was created for this Verifier (replay protection). This is usually achieved by a `nonce` and `aud` field within the Holder Binding JWT.
+       5. Check that the `typ` of the Holder Binding JWT is `hb+jwt`.
+       6. Check that the creation time of the Holder Binding JWT, as determined by the `iat` claim, is within an acceptable window.
+       7. Check that the Holder Binding JWT is valid in all other respects, per [@!RFC7519] and [@!RFC8725].
+       7. Determine that the Holder Binding JWT is bound to the current transaction and was created for this Verifier (replay protection) by validating `nonce` and `aud` claims.
```
bc-pi, line 608:
```suggestion
       7. Determine that the Holder Binding JWT is bound to the current transaction and was created for this Verifier (replay protection) by validating `nonce` and `aud` claims.
       8. Check that the Holder Binding JWT is valid in all other respects, per [@!RFC7519] and [@!RFC8725].
```

bc-pi (approved):

danielfett (approved):

PR #285, opened 2023-06-05
fix: [Hash algs] only RIPEMD versions earlier that 160 are not usable.

peppelinux:
RIPEMD-160 is usable, but not its predecessors. 

For this reason this PR removes RIPEMD-160 from the list of the weak algs by saying that only  -160 predecessors must not be used.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -719,7 +719,7 @@ alone does not indicate a hash algorithm's suitability for use in SD-JWT (it con
 heavily truncated digests, such as `sha-256-32` and `sha-256-64`, which are unfit for security
 applications).
 
-Furthermore, the hash algorithms MD2, MD4, MD5, RIPEMD-160, and SHA-1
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -719,7 +719,7 @@ alone does not indicate a hash algorithm's suitability for use in SD-JWT (it con
 heavily truncated digests, such as `sha-256-32` and `sha-256-64`, which are unfit for security
 applications).
 
-Furthermore, the hash algorithms MD2, MD4, MD5, RIPEMD-160, and SHA-1
+Furthermore, the hash algorithms MD2, MD4, MD5, SHA-1 and RIPEMD versions previous than 160,
```
bc-pi, line 722: I'd suggest that, to the extent possible, this document not try to be an arbitrator of what hash algorithms are insecure. 

```suggestion
Furthermore, the hash algorithms MD2, MD4, MD5, and SHA-1
```

bc-pi (approved):

Sakurann (approved):
Thank you!

Issue #281, opened 2023-05-30
A better name for HB-JWT is KB-JWT


danielfett:
In some use cases of SD-JWT, the key for what we currently call Holder Binding may not be bound to a holder (but, for example, to a device). We should therefore consider renaming Holder Binding to Key Binding.

bc-pi:
PRs have been merged

PR #286, opened 2023-06-06
use the term Key Binding rather than Holder Binding

bc-pi:
Change to use Key Binding and Key Binding JWT rather than Holder Binding and Holder Binding JWT. 

for issue #281

### DIFF
--- a/.github/workflows/ghpages.yml
+++ b/.github/workflows/ghpages.yml
@@ -30,6 +30,7 @@ jobs:
     - name: "Install SD-JWT tooling"
       run: |
         python3 -m pip install --upgrade pip
+        # Reminder to change below back to git+https://github.com/danielfett/sd-jwt.git before (or right after?) merging
         python3 -m pip install git+https://github.com/bc-pi/sd-jwt.git@hb2kb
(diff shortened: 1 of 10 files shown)


bc-pi:
Did some more work on this based on discussions yesterday and I believe it's now ready for review @Sakurann and @danielfett

bc-pi:
Editor's copy for this branch: https://drafts.oauth.net/oauth-selective-disclosure-jwt/bc/key-rather-than-holder/draft-ietf-oauth-selective-disclosure-jwt.html

bc-pi:
I realized/noticed that there were still things like `holder_binding` in yaml and `hb_` in file names.  So I changed `holder_binding` to `key_binding` and `hb_` to `kb_` and `holder binding` to `key binding` throughout the sd-jwt lib to try and have code, comments, and generated files reflect calling it key binding rather than holder binding. And made the corresponding changes here.

Sakurann (approved):
Thanks Brian, this looks good. I believe this is the right direction.

danielfett (approved):

Issue #241, opened 2023-03-27
Add public keys to examples


christianpaquin:
It would be useful to publish the issuer's public key used to generate the examples. This would allow implementations for validate the JWS from the spec.

danielfett:
The key pairs can be found here: https://github.com/oauth-wg/oauth-selective-disclosure-jwt/blob/master/sd_jwt/demo_settings.py

However, there is the idea of exposing all data (keys, examples, outputs of reference implementation, etc.) in a more accessible way to allow cross-library testing. I have not worked on that yet, but it's on my list.

bc-pi:
take the public key from settings.yml and put it in JWK format with some text saying it can be used to verify the examples

bc-pi:
PR https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/296 was merged

PR #296, opened 2023-06-15
add a section with public keys for validation

Sakurann:
addresses Issue #241.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1430,6 +1430,30 @@ After the validation, the Verifier will have the following data for further proc
 
 <{{examples/jsonld/verified_contents.json}}
 
+## Elliptic Curve Key Used in the Examples
+
+The following Elliptic Curve public key, represented in JWK format, can be used to validate the Issuer signatures in the above examples (with line wraps within values for display purposes only):
+
+```
+{
+  "kty": "EC",
+  "crv": "P-256",
+  "x": "b28d4MwZMjw8-00CG4xfnn9SLMVMM19SlqZpVb_uNtQ",
+  "y": "Xv5zWwuoaTgdS6hV43yI6gBwTnjukmFQQnJ_kCxzqk8"
+}
+```
+
+The following EC public key, represented in JWK format, can be used to validate the Key Binding JWT signature in the above examples (with line wraps within values for display purposes only):
+
+```
+{
+  "kty": "EC",
+  "crv": "P-256",
+  "x": "TCAER19Zvu3OHF4j4W4vfSVoHIP1ILilDls7vCeGemc",
+  "y": "ZxjiWWbZMQGHVWKVQ4hbSIirsVfuecCE6t4jT9F2HZQ"
+}
+```
+
 # Disclosure Format Considerations {#disclosure_format_considerations}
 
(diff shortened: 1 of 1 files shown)


danielfett:
The `w3c-vc` example currently uses a different key. We should call that out or change the example to the default key. I think that having one example with a different key (type) doesn't hurt, so we should call out that this example uses a different key.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1430,6 +1430,30 @@ After the validation, the Verifier will have the following data for further proc
 
 <{{examples/jsonld/verified_contents.json}}
 
+## Elliptic Curve Key Used in the Examples
+
+The following Elliptic Curve public key, represented in JWK format, can be used to validate the Issuer signatures in the above examples (with line wraps within values for display purposes only):
```
bc-pi, line 1435: This key doesn't have line wraps within values.

```suggestion
The following Elliptic Curve public key, represented in JWK format, can be used to validate the Issuer signatures in the above examples:
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1430,6 +1430,30 @@ After the validation, the Verifier will have the following data for further proc
...
+
+```
+{
+  "kty": "EC",
+  "crv": "P-256",
+  "x": "TCAER19Zvu3OHF4j4W4vfSVoHIP1ILilDls7vCeGemc",
+  "y": "ZxjiWWbZMQGHVWKVQ4hbSIirsVfuecCE6t4jT9F2HZQ"
+}
+```
+
```
bc-pi, line 1456: The holder public keys can already be found in the example content in the `cnf` claim so don't need to be here.
```suggestion

```

bc-pi (changes requested):

bc-pi:
> The `w3c-vc` example currently uses a different key. We should call that out or change the example to the default key. I think that having one example with a different key (type) doesn't hurt, so we should call out that this example uses a different key.

Either way, the holder's public key is already available in the payload of the example SD-JWTs. So it doesn't need to be and shouldn't be included here.

danielfett (approved):
LGTM with Brian's proposed changes.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1430,6 +1430,19 @@ After the validation, the Verifier will have the following data for further proc
...
+```
+{
+  "kty": "EC",
+  "crv": "P-256",
+  "x": "b28d4MwZMjw8-00CG4xfnn9SLMVMM19SlqZpVb_uNtQ",
+  "y": "Xv5zWwuoaTgdS6hV43yI6gBwTnjukmFQQnJ_kCxzqk8"
+}
+```
+
```
Sakurann, line 1445:
```suggestion

The public key used to validate a Key Binding JWT can be found in the examples as the content of the `cnf` claim.

```

Sakurann:
w3c-vc uses a different KB JWT key but the same issuer key, correct?

bc-pi:
> w3c-vc uses a different KB JWT key but the same issuer key, correct?

Yeah, i believe so. But no signed content is actually shown in that example. That example just shows the payload and disclosure content.

bc-pi (approved):

PR #301, opened 2023-06-21
Title back to "Selective Disclosure for JWTs (SD-JWT)"

bc-pi:
I've seen the "new" title from 332ded8faf20d528c5c05e04cebed28aab1a0d33 a few times in context now and it is really awkward.


Sakurann (approved):
as long as the intro text itself stays as we updated it, this makes sense to me.

bc-pi:
> as long as the intro text itself stays as we updated it, this makes sense to me.

yup, that updated intro and abstract text stays

danielfett (approved):

PR #306, opened 2023-06-23
replace `payload` with `input claim set`

Sakurann:
addressing issue #274.

- I replaced SD-JWT payload SD-JWT input claim set
- when using the word for the first time, expanded to say `The input claim set to be used as the payload of an Issuer-signed JWT component of an SD-JWT`
- kept `JSON payload` and `JWT payload` terms.

Issue #274, opened 2023-05-23
Introduce proper term for payload (before turning it into a JWT)


danielfett:
There is currently no good name for the claims before they are packaged into the SD-JWT. "Payload" can refer to the payload of the SD-JWT, which may not contain all user claims.

bc-pi:
I'm not sure what the term should be but I'd agree that text like what we've got now, "object as the set of claims that the Issuer is issuing" [from sec 5](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-04.html#section-5.2-1) isn't easy to say :)

Sakurann:
We use "the set of claims that the Issuer is issuing" when explaining in the Examples section what is in the payload. I kind of like it?

bc-pi:
"input claim set" seems to be directionality correct

Sakurann:
PR merged

PR #308, opened 2023-06-28
differentiate input claim set from the payload and clean up example structure

Sakurann:
replaces #306, addressing issue #274
I also cleaned up example structure, so that it is the same and uses the same language across multiple examples.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1547,6 +1547,7 @@ data. The original JSON data is then used by the application. See
    * Improved example for allowed variations in Disclosures
    * Added some text to the Abstract and Introduction to be more inclusive of JWS with JSON
    * Added some security considerations text about the scope of the Key Binding JWT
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -421,13 +421,11 @@ element unless a matching Disclosure for the second element is received.
 
 ## Example 1: SD-JWT {#example-1}
 
-This example uses the following object as the set of claims that the Issuer is issuing:
+In this example, the simplest SD-JWT is demonstrated.
```
bc-pi, line 424: there exists no other SD-JWT that could be considered to be more simple than this? :) 

```suggestion
In this example, a simple SD-JWT is demonstrated.
```

peppelinux: demonstrated -> demonstrated

or represented?

bc-pi: "demostrated" is not a word and "demonstrated" is what was intended

Issue #316, opened 2023-08-01
consistent terminology around "random / pseudorandom" salt


Sakurann:
Feedback received:

---
The provisions for random / pseudorandom data make use of different terminology:
•	4.1: "…a random salt…."
•	5.2.1: "…128 bits of cryptographically secure pseudorandom data…"The provisions for random / pseudorandom data make use of different terminology:
•	4.1: "…a random salt…."
•	5.2.1: "…128 bits of cryptographically secure pseudorandom data…"
•	5.6: "…cryptographically secure random number…"
•	9.3: "…cryptographically random…"
•	9.4: "…randomly-generated…"

A consistent terminology would be helpful to prevent any misinterpretations.
This terminology should not exclude pseudorandom number generators (i.e. deterministic generators) as properly designed and implemented generators are able to produce pseudorandom numbers of sufficient quality (as proven e.g. in several Common Criteria security evaluations).

•	5.6: "…cryptographically secure random number…"
•	9.3: "…cryptographically random…"
•	9.4: "…randomly-generated…"

A consistent terminology would be helpful to prevent any misinterpretations.
This terminology should not exclude pseudorandom number generators (i.e. deterministic generators) as properly designed and implemented generators are able to produce pseudorandom numbers of sufficient quality (as proven e.g. in several Common Criteria security evaluations).

Sakurann:
PR merged

PR #318, opened 2023-08-01
consistency around terminology random

Sakurann:
first pass at addressing Issue #316 and #319.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -278,8 +278,8 @@ For each claim that is an object property and that is to be made selectively dis
 
  * Create an array of three elements in this order:
    1. A salt value. MUST be a string. See (#salt-entropy) and (#salt_minlength) for security considerations. It is RECOMMENDED to base64url-encode minimum 128 bits of cryptographically secure random data, producing a string. The salt value MUST be unique for each claim that is to be selectively disclosed. The Issuer MUST NOT disclose the salt value to any party other than the Holder.
-   2. The claim name, or key, as it would be used in a regular JWS payload. The value MUST be a string.
-   3. The claim value, as it would be used in a regular JWS payload. The value MAY be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
+   2. The claim name, or key, as it would be used in a regular JWT payload. The value MUST be a string.
+   3. The claim value, as it would be used in a regular JWT payload. The value MAY be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
  * JSON-encode the array, producing an UTF-8 string.
  * base64url-encode the byte representation of the UTF-8 string, producing a US-ASCII [@RFC0020] string. This string is the Disclosure.
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -566,7 +566,7 @@ The JWT MUST contain the following elements:
 * in the JOSE header,
     * `typ`: REQUIRED. MUST be `kb+jwt`, which explicitly types the Key Binding JWT as recommended in Section 3.11 of [@!RFC8725].
     * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
-* in the JWS payload,
+* in the JWT payload,
     * `iat`: REQUIRED. The value of this claim MUST be the time at which the Key Binding JWT was issued using the syntax defined in [@!RFC7519].
     * `aud`: REQUIRED. The intended receiver of the Key Binding JWT. How the value is represented is up to the protocol used and out of scope of this specification.
     * `nonce`: REQUIRED. Ensures the freshness of the signature. The value type of this claim MUST be a string. How this value is obtained is up to the protocol used and out of scope of this specification.
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -278,8 +278,8 @@ For each claim that is an object property and that is to be made selectively dis
 
  * Create an array of three elements in this order:
    1. A salt value. MUST be a string. See (#salt-entropy) and (#salt_minlength) for security considerations. It is RECOMMENDED to base64url-encode minimum 128 bits of cryptographically secure random data, producing a string. The salt value MUST be unique for each claim that is to be selectively disclosed. The Issuer MUST NOT disclose the salt value to any party other than the Holder.
-   2. The claim name, or key, as it would be used in a regular JWT body. The value MUST be a string.
-   3. The claim value, as it would be used in a regular JWT body. The value MAY be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
+   2. The claim name, or key, as it would be used in a regular JWS payload. The value MUST be a string.
+   3. The claim value, as it would be used in a regular JWS payload. The value MAY be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
  * JSON-encode the array, producing an UTF-8 string.
  * base64url-encode the byte representation of the UTF-8 string, producing a US-ASCII [@RFC0020] string. This string is the Disclosure.
 
@@ -566,7 +566,7 @@ The JWT MUST contain the following elements:
 * in the JOSE header,
     * `typ`: REQUIRED. MUST be `kb+jwt`, which explicitly types the Key Binding JWT as recommended in Section 3.11 of [@!RFC8725].
     * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
-* in the JWT body,
+* in the JWS payload,
     * `iat`: REQUIRED. The value of this claim MUST be the time at which the Key Binding JWT was issued using the syntax defined in [@!RFC7519].
     * `aud`: REQUIRED. The intended receiver of the Key Binding JWT. How the value is represented is up to the protocol used and out of scope of this specification.
     * `nonce`: REQUIRED. Ensures the freshness of the signature. The value type of this claim MUST be a string. How this value is obtained is up to the protocol used and out of scope of this specification.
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -549,7 +549,7 @@ If the Issuer wants to enable Key Binding, it includes a public key
 associated with the Holder, or a reference thereto.
 
 It is out of the scope of this document to describe how the Holder key pair is
-established. For example, the Holder MAY provide a key pair to the Issuer,
+established. For example, the Holder MAY create a key pair and provide a public key to the Issuer,
```
Sakurann, line 552: feedback received: 
> "It is out of the scope of this document to describe how the Holder key pair is established. For example, the Holder MAY provide a key pair to the Issuer,…" Requiring the holder to provide also the private key to the issuer appears not to be a good example.

suggestion was "the Holder MAY provide a public key to the Issuer and prove possession of the corresponding private key" but I slightly played with the words.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -602,7 +602,7 @@ signed JWT. See (#enveloping) for details.
 An SD-JWT is composed of the following:
 
 * the Issuer-signed JWT
-* The Disclosures
+* optionally the Disclosures
```
Sakurann, line 605: this one might be controversial...

original feedback received was:

> According to clause 5.1 the Disclosures and the Key Binding JWT are both optional in an SD-JWT.
>
> Clause 5.11 explicitly specifies the serialized format of an SD-JWT if the Key Binding JWT is absent:
"In the case that there is no Key Binding JWT, the last
element MUST be an empty string and the last separating tilde character MUST NOT be omitted."
>
> Clause 5.11 does not explicitly specify the serialized format of an SD-JWT if the Disclosures are absent.
>
> Please specify explicitly the serialized format of an SD-JWT if the Disclosures are absent.
>
> Please indicate in the item list ("An SD-JWT is composed of the following…") that the Disclosures are optional (as already done for the Key Binding JWT).

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -730,7 +730,7 @@ In some applications or transport protocols, it is desirable to put an SD-JWT in
 
 For such use cases, the SD-JWT SHOULD be transported as a single string. Key Binding MAY be achieved by signing the envelope JWT instead of including a separate Key Binding JWT in the SD-JWT.
 
-The following non-normative example shows an SD-JWT Presentation enveloped in a JWT payload:
+The following non-normative example shows an SD-JWT Presentation enveloped in a JWS payload:
```
bc-pi, line 733: I don't totally love the suggested wording below but I think it's a "more correct" explanation of the content of the example. The text that was there previously doesn't seem quite right and JWT->JWS makes it maybe less correct. 

```suggestion
The following non-normative example shows the claims of an SD-JWT Presentation enveloped in a JWT:
```

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I'd interpreted "The Disclosures" to mean zero or more. So to me the optionality was implied.  But this seems fine too.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -277,9 +277,9 @@ Disclosures are created differently depending on whether a claim is an object pr
...
-   1. A salt value. MUST be a string. See (#salt-entropy) and (#salt_minlength) for security considerations. It is RECOMMENDED to base64url-encode minimum 128 bits of cryptographically secure pseudorandom data, producing a string. The salt value MUST be unique for each claim that is to be selectively disclosed. The Issuer MUST NOT disclose the salt value to any party other than the Holder.
-   2. The claim name, or key, as it would be used in a regular JWT body. The value MUST be a string.
-   3. The claim value, as it would be used in a regular JWT body. The value MAY be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
+   1. A salt value. MUST be a string. See (#salt-entropy) and (#salt_minlength) for security considerations. It is RECOMMENDED to base64url-encode minimum 128 bits of cryptographically secure random data, producing a string. The salt value MUST be unique for each claim that is to be selectively disclosed. The Issuer MUST NOT disclose the salt value to any party other than the Holder.
+   2. The claim name, or key, as it would be used in a regular JWS payload. The value MUST be a string.
```
bc-pi, line 281: IMHO JWT payload is fine and conveys more meaning. 
```suggestion
   2. The claim name, or key, as it would be used in a regular JWT payload. The value MUST be a string.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -277,9 +277,9 @@ Disclosures are created differently depending on whether a claim is an object pr
...
+   3. The claim value, as it would be used in a regular JWS payload. The value MAY be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
```
bc-pi, line 282:
```suggestion
   3. The claim value, as it would be used in a regular JWT payload. The value MAY be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -566,7 +566,7 @@ The JWT MUST contain the following elements:
 * in the JOSE header,
```
bc-pi, line 566:
```suggestion
* in the JWS header,
```

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: a JWS payload "can contain an arbitrary sequence of octets" while JWT conveys JSON and claims

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -566,7 +566,7 @@ The JWT MUST contain the following elements:
...
     * `typ`: REQUIRED. MUST be `kb+jwt`, which explicitly types the Key Binding JWT as recommended in Section 3.11 of [@!RFC8725].
     * `alg`: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be `none` or an identifier for a symmetric algorithm (MAC).
-* in the JWT body,
+* in the JWS payload,
```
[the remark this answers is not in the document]
bc-pi: or this would be okay by me too 
```suggestion
* in the JWT payload,
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: yeah. I agree that JWT payload is pretty well understood.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: maybe make it simpler? `the claims of an SD-JWT presentation` took me a bit of time to understand.

```suggestion
The following non-normative example shows an SD-JWT Presentation enveloped in a JWT:
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: per RFC, it's JOSE header.. https://www.rfc-editor.org/rfc/rfc7515#section-2
per DPoP/convention, JWT header might be ok too..

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: That works. It's really showing the claims/payload of that JWT in which the SD-JWT is enveloped. But that's hard to describe. And what I'd suggested is pretty bad now that I read it again.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: yeah, probably any of them would be ok actually

Sakurann:
```suggestion
* in the JWT header,
```

draft-ietf-oauth-selective-disclosure-jwt.md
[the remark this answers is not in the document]
Sakurann: yeah, let's stick to `JWT payload`.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: there are other instances where we use `JOSE header`. so let's decide in this issue comment on the term and i will change everywhere accordignly, if needed.

bc-pi: let's just keep `JOSE header`

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett:
```suggestion
* zero or more Disclosures
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -743,7 +743,7 @@ The following non-normative example shows an SD-JWT Presentation enveloped in a
 
 Here, the SD-JWT is shown as the value of an `_sd_jwt` claim where `eyJhbGci...emhlaUJhZzBZ` represents the Issuer-signed JWT and `eyJhb...dYALCGg` represents a Disclosure. The SD-JWT does not contain a Key Binding JWT as the outer container can be signed instead.
 
-Other specifications or profiles of this specification may define alternative formats for transporting an SD-JWT that envelope multiple such objects into one object, and provides Key Binding using means other than the Key Binding JWT.
+Other specifications or profiles of this specification may define alternative formats for transporting an SD-JWT that envelope multiple such objects into one object, and provide Key Binding using means other than the Key Binding JWT.
```
danielfett, line 746:
```suggestion
Other specifications or profiles of this specification may define alternative formats for transporting an SD-JWT that envelope multiple such objects into one object and provide Key Binding using means other than the Key Binding JWT.
```

danielfett (approved):
Approved with the proposed changes.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I like the Dr.'s suggestion above of "zero or more Disclosures"

bc-pi (approved):

Issue #333, opened 2023-08-25
claim validation should happen after processing disclosures


bc-pi:
Validation of JWT claims, especially those controlling the validity such as iss, exp, or nbf, needs to occur after the processing/unpacking of the disclosures. Such claims will likely usually be in plain text of the payload but as written now https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#section-6.1-4.2.2.4 checks on them could be omitted, if they are in disclosures. [as discussed after OSW in the Uber to LHR]

PR #334, opened 2023-08-28
Ensure claims that control validity are checked after decoding payload

danielfett:
This fixes Issue #333.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -264,7 +264,7 @@ The payload of an SD-JWT is a JSON object according to the following rules:
 
 Applications of SD-JWT SHOULD be explicitly typed using the `typ` header parameter. See (#explicit_typing) for more details.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -682,6 +681,7 @@ an SD-JWT:
     5. Remove all array elements for which the digest was not found in the previous step.
     6. Remove all `_sd` keys and their contents from the Issuer-signed JWT payload.
     7. Remove the claim `_sd_alg` from the SD-JWT payload.
+ 4. Check that the SD-JWT is valid using the `nbf`, `iat`, and `exp` claims in the processed payload. If a required validity-controlling claim is missing (see (#sd-validity-claims)), the SD-JWT MUST be rejected.
```
bc-pi, line 684:
```suggestion
 4. Check that the SD-JWT is valid using claims such as `nbf`, `iat`, and `exp` in the processed payload. If a required validity-controlling claim is missing (see (#sd-validity-claims)), the SD-JWT MUST be rejected.
```

bc-pi (approved):

Sakurann (approved):

Issue #337, opened 2023-09-07
Example 4a - address array disclosure


sschulz-t:
I was wondering if there might be a mistake in the encoding of the address given in example 4:
```
<snip>
"birthdate": "1973-01-01",
  "address": {
    "postal_code": "12345",
    "locality": "Irgendwo",
    "street_address": "Sonnenstrasse 23",
    "country_code": "DE"
  },
  "is_over_18": true,
<snip>
```

The documentation says:
```
SHA-256 Hash: Y1urWJV_-HBGnSf9tFOwvH4cICRBCiKwEHfkXFSfjpo
Disclosure:
WyItNmhDZVJFV3JjRWg0Q2JzQ2RjVTVRIiwgImFkZHJlc3MiLCB7Il9zZCI6
IFsiOHo4ejlYOWpVdGI5OWdqZWpDd0ZBR3o0YXFsSGYtc0NxUTZlTV9xbXBV
USIsICJDeHE0ODcyVVhYbmdHVUxUX2tsOGZkd1ZGa3lLNkFKZlBaTHk3TDVf
MGtJIiwgImxjMk8weER4WXdiZ2M4cjJERXc3eWhfSURXZXhXOENiZ3R6WVBR
RlJpNGMiXSwgInBvc3RhbF9jb2RlIjogIjEyMzQ1IiwgImxvY2FsaXR5Ijog
IklyZ2VuZHdvIiwgInN0cmVldF9hZGRyZXNzIjogIlNvbm5lbnN0cmFzc2Ug
MjMiLCAiY291bnRyeV9jb2RlIjogIkRFIn1d
Contents: ["-6hCeREWrcEh4CbsCdcU5Q", "address", {"_sd":
["8z8z9X9jUtb99gjejCwFAGz4aqlHf-sCqQ6eM_qmpUQ",
"Cxq4872UXXngGULT_kl8fdwVFkyK6AJfPZLy7L5_0kI",
"lc2O0xDxYwbgc8r2DEw7yh_IDWexW8CbgtzYPQFRi4c"],
"postal_code": "12345", "locality": "Irgendwo",
"street_address": "Sonnenstrasse 23", "country_code": "DE"}]
```

I assume this should be [SD-JWT with Recursive Disclosures](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#name-option-3-sd-jwt-with-recurs)?

[...]

PR #338, opened 2023-09-07
Clean up Example 4a by not including decoys to fix issue #337

bc-pi:


danielfett (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1372,10 +1373,6 @@ The following Disclosures are created by the Issuer:
 
 {{examples/arf-pid/disclosures.md}}
 
-The following decoy digests are added:
-
-{{examples/simple_structured/decoy_digests.md}}
```
bc-pi, line 1377: Note that this was pointing to the wrong file...

Sakurann: why pointing to a correct file not an option?

bc-pi: This example doesn't use decoys anymore (probably shouldn't have in the first place) so there's no decoy digests file content to point to. #337 was extremely confusing to get to the bottom of due in large part to including the decoy_digests.md from a different example.

Sakurann (approved):
I don't know why but I think we should have more examples, but not going to go against these changes.

Issue #320, opened 2023-08-02
W3C VC Example lacks cnf


OR13:
https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#appendix-A.4

compared to https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#section-5.5

expected to see this:

```json
{
  "iss": "https://example.com/issuer",
  "iat": 1683000000,
  "exp": 1883000000,
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/vaccination/v1"
  ],
  "type": [
    "VerifiableCredential",
    "VaccinationCertificate"
  ],
  "issuer": "https://example.com/issuer",
  "issuanceDate": "2023-02-09T11:01:59Z",
  "expirationDate": "2028-02-08T11:01:59Z",
  "name": "COVID-19 Vaccination Certificate",
  "description": "COVID-19 Vaccination Certificate",
  "credentialSubject": {
    "vaccine": {
      "type": "Vaccine",
      "atcCode": "J07BX03",
      "medicinalProductName": "COVID-19 Vaccine Moderna"
    },
    "recipient": {
      "type": "VaccineRecipient"
    },
    "type": "VaccinationEvent",
    "order": "3/3",
    "dateOfVaccination": "2021-06-23T13:40:12Z"
  },
  "_sd_alg": "sha-256",
  // this part needs to be added.
  "cnf": {
    "jwk": {
      "kty": "EC",
      "crv": "P-256",
      "x": "TCAER19Zvu3OHF4j4W4vfSVoHIP1ILilDls7vCeGemc",
      "y": "ZxjiWWbZMQGHVWKVQ4hbSIirsVfuecCE6t4jT9F2HZQ"
    }
  }
}
```

alenhorvat:
VC can be signed with JWS only without JWT claims or cnf.
I guess both examples could/should be presented. One would be securing using JWS, other would be securing using JWT.

OR13:
@alenhorvat your comment seems not relevant to the issue... the issue is that the text suggests `cnf` will be present in the example, but it is not. 

> VC can be signed with JWS only without JWT claims or cnf.

In w3c vcdm v1.1 `typ` MUST be JWT.
In w3c vcdm v2, see https://github.com/w3c/vc-jose-cose

[...]

PR #339, opened 2023-09-07
Add cnf claim to Example 4b (per issue #320)

bc-pi:


OR13 (approved):

Sakurann (changes requested):
description of example 4b also needs to be updated to say the example contains key binding

bc-pi:
> description of example 4b also needs to be updated to say the example contains key binding

It does already say "Key Binding is applied using the Holder's public key passed in a cnf claim in the SD-JWT" right there at https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#appendix-A.4 which is, I think, part of why Orie filed #320 asking that the example content reflect that statement by having a cnf.

OR13:
Also for context, we are trying to use whatever examples you folks define in the official WG test suite:

https://github.com/w3c/vc-jose-cose-test-suite/blob/main/testcases/secured-vc-jwt-sd/spec.yaml

Sakurann (approved):
I might have misunderstood, sorry

Issue #290, opened 2023-06-13
Improve Structure, especially around Example 1


danielfett:
_Originally posted by @Sakurann in https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/283#discussion_r1227387227_

>               I don't disagree this section probably belongs here in terms of the order, but the level of this section feels off (looked at [the rendered ver of this branch](https://drafts.oauth.net/oauth-selective-disclosure-jwt/danielfett/arrays/draft-ietf-oauth-selective-disclosure-jwt.html#name-format-of-an-sd-jwt)).
> 
> I think there is a logical separations between sections from 5.1 to 5.5 (basic operations) and from sections 5.6 to 5.7 (advanced concepts) from 5.8 to 5.9 (content of the SD-JWT) and 5.10 to 5.11 (combined formats). and having them all at the same level makes it hard to read..

I agree. There will be other changes coming to the presentation/issuance format sections and a new section for the JWS JSON Serialization. After these changes, we should revisit the current structure and modify as needed.

Sakurann:
@rohan-wire, were you planning to do a PR that could partially address this?

Sakurann:
PR merged

PR #347, opened 2023-10-04
Improve structure of specification

danielfett:
To fix Issue #290, this PR reorganizes what was previously section 5 as follows:

Screenshot_20231004_144710

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -286,7 +286,7 @@ MUST NOT use `none` or an identifier for a symmetric algorithm (MAC).
 The payload of an SD-JWT is a JSON object according to the following rules:
 
  1. The payload MAY contain the `_sd_alg` key described in (#hash_function_claim).
- 2. The payload MAY contain one or more digests of Disclosures to enable selective disclosure of the respective claims, created and formatted as described below.
+ 2. The payload MAY contain one or more digests of Disclosures to enable selective disclosure of the respective claims, created and formatted as described in (#creating_disclosures).
  3. The payload MAY contain one or more decoy digests to obscure the actual number of claims in the SD-JWT, created and formatted as described in (#decoy_digests).
  4. The payload MAY contain one or more non-selectively disclosable claims.
  5. The payload MAY also contain Holder's public key(s) or reference(s) thereto, as well as further claims such as `iss`, `iat`, etc. as defined or required by the application using SD-JWTs. See (#holder_public_key_claim) for more details.
@@ -299,6 +299,38 @@ It is the Issuer who decides which claims are selectively disclosable and which 
 
 Claims that are not selectively disclosable are included in the SD-JWT in plaintext just as they would be in any other JSON structure.
 
+
+### Hash Function Claim {#hash_function_claim}
+
+The claim `_sd_alg` indicates the hash algorithm used by the Issuer to generate
+the digests as described in (#creating_disclosures). When used, this claim MUST
+appear at the top level of the SD-JWT payload. It
+MUST NOT be used in any object nested within the payload. If the  `_sd_alg`
+claim is not present at the top level, a default value of `sha-256` MUST be used.
+
+The hash algorithm identifier MUST be a hash algorithm value from the "Hash Name
+String" column in the IANA "Named Information Hash Algorithm" registry
+[@IANA.Hash.Algorithms] or a value defined in another specification and/or
+profile of this specification.
+
+To promote interoperability, implementations MUST support the `sha-256` hash
+algorithm.
+
+See (#security_considerations) for requirements regarding entropy of the salt,
+minimum length of the salt, and choice of a hash algorithm.
+
+### Holder Public Key Claim {#holder_public_key_claim}
+
+If the Issuer wants to enable Key Binding, it includes a public key
+associated with the Holder, or a reference thereto.
+
+It is out of the scope of this document to describe how the Holder key pair is
+established. For example, the Holder MAY create a key pair and provide a public key to the Issuer,
+the Issuer MAY create the key pair for the Holder, or
+Holder and Issuer MAY use pre-established key material.
+
+Note: The examples in this document use the `cnf` claim defined in [@RFC7800] to include
+the raw public key by value in SD-JWT.
 
 ## Disclosures {#creating_disclosures}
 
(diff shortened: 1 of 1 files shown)


danielfett:
Note: I also changed the story in Section 5 from a bottom-up to a top-down approach (structure of compact serialization SD-JWT first, then going into the details).

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -242,11 +242,42 @@ At a high level, the Verifier
...
+* an Issuer-signed JWT,
+* zero or more Disclosures, and
+* optionally a Key Binding JWT.
+
+The individual parts will be explained in the following subsections.
+
+The serialized format for the SD-JWT is the concatenation of each part delineated with a single tilde ('~') character as follows:
+
+```
+<Issuer-signed JWT>~<Disclosure 1>~<Disclosure 2>~...~<Disclosure N>~<optional KB-JWT>
```
bc-pi, line 258:
```suggestion
<Issuer-signed JWT>~<Disclosure 1>~<Disclosure 2>~...~<Disclosure N>~<optional KB-JWT>

```

adding a newline so (hopefully) the horizontal scrollbar in the HTML output doesn't cover up the content

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -422,9 +453,47 @@ disclosable array elements for which they did not receive a Disclosure. In the
...
+
+Below is a non-normative example of a Key Binding JWT header:
+
+```
+{
+  "alg": "ES256",
+  "typ": "kb+jwt"
+}
+```
+
```
bc-pi, line 482: With the other examples gone from this section - this one feels a bit out of place. And not needed. 
```suggestion
```

bc-pi (commented):
This looks pretty good and does generally improve the structure. 
The Decoy Digests, Hash Function Claim, Holder Public Key Claim subsections feel a bit out of place in section 7.  What would you think about moving them to subsections under 5.1.  Issuer-signed JWT Payload? The ToC might look something like this:

Screenshot 2023-10-04 at 3 08 50 PM

bc-pi (changes requested):
I'd like to see a few more subsections moved https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/347#pullrequestreview-1658506950 and one minor example removed https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/347/files#r1346423276 but otherwise this looks good.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -242,11 +242,43 @@ At a high level, the Verifier
...
+<Issuer-signed JWT>~<Disclosure 1>~<Disclosure 2>~...~<Disclosure N>~<optional KB-JWT>
+
+```
+
+The order of the tilde separated values MUST be the Issuer-signed JWT, followed by any number of Disclosures, and lastly the optional Key Binding JWT.
+In the case that there is no Key Binding JWT, the last element MUST be an empty string and the last separating tilde character MUST NOT be omitted.
 
-## SD-JWT Payload
+The Disclosures are linked to the SD-JWT payload through the
+digest values included therein.
```
Sakurann, line 266:
```suggestion
The Disclosures are linked to the Issuer-signed JWT by including
Disclosure digest values therein.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -242,11 +242,43 @@ At a high level, the Verifier
...
+
+When issued to a Holder, the Issuer includes all the relevant Disclosures in the SD-JWT.
+
+For presentation to a Verifier, the Holder sends the SD-JWT including only its selected
+set of the Disclosures to the Verifier.
```
Sakurann, line 271:
```suggestion
When issuing to a Holder, the Issuer includes all the relevant Disclosures in the SD-JWT.

When presenting to a Verifier, the Holder sends only the selected set of the Disclosures in the SD-JWT.
```
consistency here felt better?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -242,11 +242,43 @@ At a high level, the Verifier
...
+
+The Holder MAY send any subset of the Disclosures to the Verifier, i.e.,
+none, multiple, or all Disclosures. For data that the Holder does not want to reveal
```
Sakurann, line 274:
```suggestion
none, parts of, or all Disclosures. For data that the Holder does not want to reveal
```
`parts of ` feels a little off, but `all` is part of `multiple`..

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -363,14 +395,14 @@ The SHA-256 digest of the Disclosure
 `WyJsa2x4RjVqTVlsR1RQVW92TU5JdkNBIiwgIkZSIl0` would be
 `w0I8EKcdCtUPkGCNUrfwVp2xEgNjtoIDlOxc9-PlOhs`.
 
-## Embedding Disclosure Digests in SD-JWTs {#embedding_disclosure_digests}
+### Embedding Disclosure Digests in SD-JWTs {#embedding_disclosure_digests}
 
 For selectively disclosable claims, the digests of the Disclosures are embedded into the SD-JWT instead of the claims themselves. The precise way of embedding depends on whether a claim is an object property (key-value pair) or an array element.
```
Sakurann, line 400:
```suggestion
For selectively disclosable claims, the digests of the Disclosures are embedded into the Issuer-signed JWT instead of the claims themselves. The precise way of embedding depends on whether a claim is an object property (key-value pair) or an array element.
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: should this example be re-added to another section? it was added because people wanted to see an example of a KB-JWT header..

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -451,25 +521,25 @@ The payload is then signed by the Issuer to create a JWT like the following:
...
-To ensure readability and replicability, the examples in this specification do
-not contain decoy digests unless explicitly stated. For an example
-with decoy digests, see (#example-simple_structured).
+<{{examples/simple/sd_jwt_presentation.txt}}
+
+The following Key Binding JWT payload was created and signed for this presentation by the Holder:
+
+<{{examples/simple/kb_jwt_payload.json}}
+
+# Data Format Details
```
Sakurann, line 542: it almost feels like `# Advanced Features`

Sakurann:
> The Decoy Digests, Hash Function Claim, Holder Public Key Claim subsections feel a bit out of place in section 7. What would you think about moving them to subsections under 5.1. Issuer-signed JWT Payload?

+1 to this

Sakurann (changes requested):
reads so much better and flows more naturally - agree with Brian's suggested subsections 5.1.1~5.1.3. also made few suggestions

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: `some of` ?

danielfett:
```suggestion
none, some, or all Disclosures. For data that the Holder does not want to reveal
```
WDYT?

bc-pi: WFM

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: I agree that removing makes most sense. We're not showing any other headers in the document, there is nothing surprising in this header, and people can still decode the KB-JWT if they like to see a header.

danielfett:
I made the additional changes proposed by Brian, with the following deltas:

* I moved the section "Decoy Digests" to be the last subsection of "5.2 Disclosures". It did not make sense in 5.1 as not all concepts were introduced there yet.
* I slightly reformatted the list of requirements on the SD-JWT payload in Section 5.1.
* I noticed that the note on line breaks in the examples is not in the correct place and moved it towards the beginning of Section 6 (Example 1).

bc-pi:
> I moved the section "Decoy Digests" to be the last subsection of "5.2 Disclosures". It did not make sense in 5.1 as not all concepts were introduced there yet. 

That makes sense and is better than what I'd suggested.

bc-pi (approved):

OR13 (approved):

Issue #315, opened 2023-07-13
JWT claims IANA registration requests


bc-pi:
This [TBD](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#section-12-1) was meant as a placeholder/reminder for JWT Claims Registration requests to IANA that would be needed eventually.   

 The `_sd` and `_sd_alg` claims should be registered. 

I'm not sure about `...` to be honest, which is ironic b/c I'm supposed to know about such things (see the so-called experts [listed in the registry](https://www.iana.org/assignments/jwt/jwt.xhtml#claims)). But it's somewhat different than the other claims in the registry. But, I dunno, requesting registration is probably the way to go w/ `...`.

Sakurann:
I share your unsureness, and agree that requesting registration is the way to go.
btw in the spec, we have a text registered `...` claim, so if we decide not to register it, that text needs to be updated.

bc-pi:
> btw in the spec, we have a text registered `...` claim

Wait... really, where?

danielfett:
PR has been merged.

PR #348, opened 2023-10-05
Added JWT claims registration requests to IANA

bc-pi:
Added JWT claims registration requests to IANA (to fix issue #315)

https://drafts.oauth.net/oauth-selective-disclosure-jwt/bc-IANA-JWT/draft-ietf-oauth-selective-disclosure-jwt.html#name-json-web-token-claims-regis

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1128,7 +1128,7 @@ IANA "JSON Web Token Claims" registry [@IANA.JWT] established by [@!RFC7519].
 <br/>
 
 *  Claim Name: `...`
-*  Claim Description: Array element disclosure digest
+*  Claim Description: Digest of the Disclosure for an array element
 *  Change Controller: IETF
 *  Specification Document(s):  [[ (#embedding_array_elements) of this specification ]]
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1119,21 +1119,20 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
 This specification requests registration of the following Claims in the
 IANA "JSON Web Token Claims" registry [@IANA.JWT] established by [@!RFC7519].
 
-Object Property Disclosure Digests:
 
 *  Claim Name: `_sd`
 *  Claim Description: Digests of Disclosures for object properties
 *  Change Controller: IETF
 *  Specification Document(s):  [[ (#embedding_object_properties) of this specification ]]
 
-Array Element Disclosure Digest:
+<br/>
 
 *  Claim Name: `...`
 *  Claim Description: Array element disclosure digest
 *  Change Controller: IETF
 *  Specification Document(s):  [[ (#embedding_array_elements) of this specification ]]
 
-Disclosure Digest Hash Function:
+<br/>
 
 *  Claim Name: `_sd_alg`
 *  Claim Description: Hash algorithm used to generate disclosure digests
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1114,7 +1114,31 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
 
 # IANA Considerations {#iana_considerations}
 
-TBD
+## JSON Web Token Claims Registration
+
+This specification requests registration of the following Claims in the
+IANA "JSON Web Token Claims" registry [@IANA.JWT] established by [@!RFC7519].
+
+Object Property Disclosure Digests:
```
Sakurann, line 1122: do we need these headers `Object Property Disclosure Digests:`, `Array Element Disclosure Digest:`, `Disclosure Digest Hash Function:`..? they are the same as `claim description` values rn.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1114,7 +1114,31 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
...
+
+*  Claim Name: `_sd`
+*  Claim Description: Object property disclosure digests
```
Sakurann, line 1125:
```suggestion
*  Claim Description: Disclosure digests
```
I am sorry, i could not understand `Object property disclosure digests`..? did you want to say that digests are hashes over objects?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1114,7 +1114,31 @@ The work on this draft was started at OAuth Security Workshop 2022 in Trondheim,
...
+*  Change Controller: IETF
+*  Specification Document(s):  [[ (#embedding_object_properties) of this specification ]]
+
+Array Element Disclosure Digest:
+
+*  Claim Name: `...`
+*  Claim Description: Array element disclosure digest
```
Sakurann, line 1132:
```suggestion
*  Claim Description: Digests of Disclosures for array elements
```

Sakurann (changes requested):
made suggestions to the claim descriptions as i was confused..

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: They are there to breakup the sets of bullets for each claim that run together without something. And I honestly don't know how else to do it with the markdown to xml to html/txt/etc tool chain in place. 

So we kinda need something but it doesn't have to be these exactly.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: sure, if that's more understandable

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: ok but it would be singular

bc-pi: so how about this?
```suggestion
*  Claim Description: Digest of the Disclosure for an array element
```

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I just saw this example https://www.rfc-editor.org/rfc/rfc7515#section-9.1.2, but yeah.
maybe use claim names as the `breakups`?

bc-pi: actually, maybe `<br/>` can be used to breakup with no text needed ....

bc-pi: that seems to work ok

Sakurann (approved):

danielfett (approved):

Issue #327, opened 2023-08-09
Request a media type for JSON Serialization


OR13:
https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#section-12.1

Something like `+sd-jwt+json`, that supports the JSON serialization and in particular the unprotected header.

alenhorvat:
After some reflection I realised that SD-JWT (compact/JSON) is not a new encoding for securing mechanism but it is a presentation since it only adds information to existing JWS (compact/JSON). In a sense it is an alternative to Verifiable Presentation.

I guess
`+sd-jwt`
`+sd-jwt+json`
cover the two cases.

Sakurann:
per [IANA registry](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml), `the syntax and semantics of fragment identifiers specified for +json SHOULD be as specified for "application/json". [[RFC8259](https://www.rfc-editor.org/rfc/rfc8259.html)]` 

`+json` has nothing to do with JWS JSON serialization. and I don't think media type has been the mechanism used to detect JWS serialization format.

Sakurann:
I do not see the need to request registration of a new media type. `+sd-jwt` and `+json` will both exist independently and if someone sees the need to combine they could but I do not feel comfortable advocating for the use of combining them. It does not have to be a media type that is used to differentiate JWS compact and JSON serializations.

alenhorvat:
I'm just trying to follow the JWS RFC:

_The "typ" value "JOSE" can be used by applications to indicate that
   this object is a JWS or JWE using the JWS Compact Serialization or
   the JWE Compact Serialization.  The "typ" value "JOSE+JSON" can be
   used by applications to indicate that this object is a JWS or JWE
   using the JWS JSON Serialization or the JWE JSON Serialization.
   Other type values can also be used by applications._

Since the two encodings are completely different, we should be able to communicate the exact encoding, else devs need to try whether it is compact or JSON.

OR13:
@Sakurann do you see a reason for `+jwt` to be here: https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml

but for `+sd-jwt` to not be?

won't that just lead to `application/vc+sd-jwt` using an "unregistered structured suffix" ?

`application/sd-jwt+json` would mirror `application/jose+json`

[...]

PR #349, opened 2023-10-05
Request a media type for the JSON Serialization

bc-pi:
for #327

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1142,7 +1142,30 @@ To indicate that the content is an SD-JWT:
 * Intended usage: COMMON
 * Restrictions on usage: none
 * Author: Daniel Fett, mail@danielfett.de
-* Change Controller: IESG
+* Change Controller: IETF
+* Provisional registration?  No
+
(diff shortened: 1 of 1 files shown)


OR13 (approved):

Sakurann (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1135,16 +1135,41 @@ To indicate that the content is an SD-JWT:
...
 * Person & email address to contact for further information: Daniel Fett, mail@danielfett.de
 * Intended usage: COMMON
 * Restrictions on usage: none
 * Author: Daniel Fett, mail@danielfett.de
-* Change Controller: IESG
+* Change Controller: IETF
 * Provisional registration?  No
 
+<br/>
+To indicate that the content is a JWS JSON serialized SD-JWT:
```
Sakurann, line 1149: to be consistent and bring clarity, we could change line 1125 to `To indicate that the content is a JWS compact serialized SD-JWT:` but Brian convinced me that no need because we do not really use the term compact serialized in the text and it is kind of obvious without the text changes I considered.

danielfett (approved):

Issue #342, opened 2023-09-22
mention that key binding key can be correlatable in batching 


bc-pi:
It might be obvious to some/many but I think it's worth mentioning that the key binding key needs to be unique per SD-JWT when using batch issuance to try and get verifier/verifier unthinkably. I feel like I've seen it come up in a few discussions recently and sometimes stated like it was a new discovery. Having just a mention of it in the draft could be helpful guidance/reminder and also preempt criticism that it was somehow not known. 

Maybe just a few words in https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#name-unlinkability

Sakurann:
making it clearer would be nice

paulbastian:
In batch issued credentials it might make sense to also think about iat/exp dates
Either these should be:
- round up/down to certain timepoints, e.g. 24h UTC times giving a better herd anonymity
- randomize the timepoints within a certain range

Sakurann:
PR closed

PR #354, opened 2023-10-17
Rewrite unlinkability considerations

danielfett:
Addresses Issue #342 

Instead of just adding text re the key binding keys, I rewrote the whole section, as the previous one was more of a placeholder.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1148,7 +1148,7 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
 ## Unlinkability
 
 Unlinkability is a property whereby adversaries are prevented from correlating
-credential presentations of the same user (beyond the user's consent).
+credential presentations of the same user beyond the user's consent.
 Otherwise, an adversary might be able to learn more about the user than the user
 intended to disclose, for example:
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1079,32 +1079,61 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
 
 ## Unlinkability
 
-Unlinkability is a property that prevents an adversary from associating two or
-more sessions of the same user. Two types of unlinkability are often considered:
+Unlinkability is a property that prevents an adversary from correlating
+credential presentations of the same user (beyond the user's consent).
+Otherwise, an adversary might be able to learn more about the user than the user
+intended to disclose, for example:
 
-- Verifier/Verifier unlinkability: Two colluding Verifiers should not be able to
-  tell that they have received presentations from the same user.
-- Issuer/Verifier unlinkability: An Issuer colluding with a Verifier should
-  not be able to tell that a user that has received a credential from the Issuer
-  has presented it to the Verifier.
+ * Cooperating Verifiers might want to track users across services to build
+   advertising profiles.
+ * Issuers might want to track where users present their credentials to enable
+   surveillance.
+ * After a data breach at multiple Verifiers, publicly available information
+   might allow linking identifiable information presented to Verifier A with
+   originally anonymous information presented to Verifier B, therefore revealing
+   the identities of users of Verifier B.
 
-In both cases, unlinkability is limited to cases where the
-disclosed claims do not contain information that directly or indirectly
-identifies the user. For example, when a tax ID is contained in the disclosed
-claims, the Issuer and Verifier can easily link the user's sessions. However,
-when the user only discloses a birthdate to one Verifier and a postal code to
-another Verifier, the two Verifiers should not be able to tell that they were
-talking to the same user.
+The following types of unlinkability are considered here:
 
-Issuer/Verifier unlinkability cannot be achieved in salted-hash based approaches to credentials, such as SD-JWT, as the issued
-credential is directly forwarded to the Verifier.
+- Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
+  learn that they have received presentations from the same user.
+- Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
+  should not be able to tell that a user presented the credential to a certain
+  Verifier (that is behaving according to protocol).
+- Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
+  credential should not be able to tell that a user presented the credential to
+  a certain Verifier (even if the Verifier colludes with the Issuer or becomes
+  compromised and leaks stored credentials from presentations).
+
+In all cases, unlinkability is limited to cases where the disclosed claims do
+not contain information that directly or indirectly identifies the user. For
+example, when a tax ID is contained in the disclosed claims, the Issuer and
+Verifier can easily link the user's transactions. However, when the user only
+discloses a birthdate to one Verifier and a postal code to another Verifier, the
+two Verifiers should not be able to tell that they were talking to the same
+user.
+
+Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be
+achieved in salted-hash based approaches to credentials, such as SD-JWT, as the
+issued credential is directly forwarded to the Verifier, who can forward it to
+the malicious Issuer.
+
+Issuer/Verifier unlinkability with an honest Verifier can be achieved. A
+'phone-home' from the Verifier to the Issuer, for example, for a revocation
+check, SHOULD be avoided, or, if that is not possible, MUST be done in a
+privacy-preserving way where details about the credential are not revealed to
+the Issuer. (Even then, timing of such requests might remain as a side-channel.)
 
 Verifier/Verifier unlinkablility can be achieved using batch issuance: A batch
 of credentials based on the same claims is issued to the Holder instead of just
 a single credential. The Holder can then use a different credential for each
 Verifier or even for each session with a Verifier. New key binding keys and
 salts MUST be used for each credential in the batch to ensure that the Verifiers
-cannot link the credentials using these values.
+cannot link the credentials using these values. Likewise, claims carrying time
+information, like `iat`, `exp`, and `nbf`, MUST either be randomized within a
+time period considered appropriate (e.g., randomize `iat` within the last 24
+hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
+beginning of the day).
 
 ## Issuer Identifier
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1193,7 +1193,7 @@ Where such callbacks are necessary, they MUST be executed in a manner that
 preserves privacy and does not disclose details about the credential to the Issuer. It is
 important to note that the timing of such requests could potentially serve as a side-channel.
 
-Verifier/Verifier unlinkablility can be achieved using batch issuance: A batch
+Verifier/Verifier unlinkablility and presentation unlinkablility can be achieved using batch issuance: A batch
 of credentials based on the same claims is issued to the Holder instead of just
 a single credential. The Holder can then use a different credential for each
 Verifier or even for each session with a Verifier. New key binding keys and
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1147,7 +1147,7 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
 
 ## Unlinkability
 
-Unlinkability is a property that prevents an adversary from correlating
+Unlinkability is a property whereby adversaries are prevented from correlating
 credential presentations of the same user (beyond the user's consent).
 Otherwise, an adversary might be able to learn more about the user than the user
 intended to disclose, for example:
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
 
 ## Unlinkability
 
-Colluding Issuer/Verifier or Verifier/Verifier pairs could link issuance/presentation
-or two presentation sessions to the same user on the basis of unique values encoded in the SD-JWT
-(Issuer signature, salts, digests, etc.).
+Unlinkability is a property that prevents an adversary from associating two or
+more sessions of the same user. Two types of unlinkability are often considered:
 
-To prevent these types of linkability, various methods, including but not limited to the following ones can be used:
+- Verifier/Verifier unlinkability: Two colluding Verifiers should not be able to
+  tell that they have received presentations from the same user.
+- Issuer/Verifier unlinkability: An Issuer colluding with a Verifier should
+  not able to tell that a user that has received a credential from the Issuer
+  has presented it to the Verifier.
 
-- Use advanced cryptographic schemes, outside the scope of this specification.
-- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
+In both cases, unlinkability is in principle limited to cases where the
+disclosed claims do not contain information that directly or indirectly
+identifies the user. For example, when a tax ID is contained in the disclosed
+claims, the Issuer and Verifier can easily link the user's sessions. However,
+when the user only discloses a birthdate to one Verifier and a postal code to
+another Verifier, the two Verifiers should not be able to tell that they were
+talking to the same user.
 
+With SD-JWT, Issuer/Verifier unlinkablility cannot be achieved, as the issued
+SD-JWT is directly forwarded to the Verifier.
+
+Verifier/Verifier unlinkablility can be achieved using batch issuance: A batch
+of credentials based on the same claims is issued to the Holder instead of just
+a single credential. The Holder can then use a different credential for each
+Verifier or even for each session with a Verifier. New key binding keys and
+salts MUST be used for each credential in the batch to ensure that the Verifiers
+cannot link the credentials using these values.
 
 ## Issuer Identifier
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1149,7 +1149,7 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
 
 Unlinkability is a property whereby adversaries are prevented from correlating
 credential presentations of the same user beyond the user's consent.
-Otherwise, an adversary might be able to learn more about the user than the user
+Without unlinkability, an adversary might be able to learn more about the user than the user
 intended to disclose, for example:
 
  * Cooperating Verifiers might want to track users across services to build
@@ -1186,7 +1186,7 @@ the malicious Issuer.
 
 Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
 However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
-disclose information about the credential's usage to the Verifier. Therefore, such callbacks
+disclose information about the credential's usage to the Issuer. Therefore, such callbacks
 SHOULD be avoided. Where they are unavoidable, they MUST be executed in a manner that
 preserves privacy and does not disclose details about the credential to the Issuer. It is
 important to note that the timing of such requests could potentially serve as a side-channel.
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1167,11 +1167,11 @@ The following types of unlinkability are considered here:
    learn that they have received presentations from the same user.
  * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
    should not be able to know that a user presented the credential to a certain
-   Verifier (that is behaving according to protocol).
+   Verifier that is behaving according to protocol.
  * Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
    credential should not be able to tell that a user presented the credential to
-   a certain Verifier (even if the Verifier colludes with the Issuer or becomes
-   compromised and leaks stored credentials from presentations).
+   a certain Verifier, even if the Verifier colludes with the Issuer or becomes
+   compromised and leaks stored credentials from presentations.
 
 In all cases, unlinkability is limited to cases where the disclosed claims do
 not contain information that directly or indirectly identifies the user. For
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1186,8 +1186,8 @@ the Issuer.
 
 Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
 However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
-disclose information about the credential's usage to the Issuer. Therefore, such callbacks
-SHOULD be avoided. Where they are unavoidable, they MUST be executed in a manner that
+disclose information about the credential's usage to the Issuer.
+Where such callbacks are necessary, they MUST be executed in a manner that
 preserves privacy and does not disclose details about the credential to the Issuer. It is
 important to note that the timing of such requests could potentially serve as a side-channel.
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1095,15 +1095,15 @@ intended to disclose, for example:
 
 The following types of unlinkability are considered here:
 
-- Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
-  learn that they have received presentations from the same user.
-- Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
-  should not be able to tell that a user presented the credential to a certain
-  Verifier (that is behaving according to protocol).
-- Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
-  credential should not be able to tell that a user presented the credential to
-  a certain Verifier (even if the Verifier colludes with the Issuer or becomes
-  compromised and leaks stored credentials from presentations).
+ * Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
+   learn that they have received presentations from the same user.
+ * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
+   should not be able to tell that a user presented the credential to a certain
+   Verifier (that is behaving according to protocol).
+ * Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
+   credential should not be able to tell that a user presented the credential to
+   a certain Verifier (even if the Verifier colludes with the Issuer or becomes
+   compromised and leaks stored credentials from presentations).
 
 In all cases, unlinkability is limited to cases where the disclosed claims do
 not contain information that directly or indirectly identifies the user. For
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1163,8 +1163,10 @@ intended to disclose, for example:
 
 The following types of unlinkability are considered here:
 
+ * Presentation Unlinkability: A Verifier should not be able to link two
+   presentations of the same credential.
  * Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
-   learn that they have received presentations from the same user.
+   learn that they have received presentations of the same credential.
  * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
    should not be able to know that a user presented the credential to a certain
    Verifier that is not behaving maliciously.
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1184,11 +1184,12 @@ achieved in salted-hash based approaches to credentials, such as SD-JWT, as the
 issued credential is directly forwarded to the Verifier, who can forward it to
 the malicious Issuer.
 
-Issuer/Verifier unlinkability with an honest Verifier can be achieved. A
-'phone-home' from the Verifier to the Issuer, for example, for a revocation
-check, SHOULD be avoided, or, if that is not possible, MUST be done in a
-privacy-preserving way where details about the credential are not revealed to
-the Issuer. (Even then, timing of such requests might remain as a side-channel.)
+Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
+However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
+disclose information about the credential's usage to the Verifier. Therefore, such callbacks
+SHOULD be avoided. Where they are unavoidable, they MUST be executed in a manner that
+preserves privacy and does not disclose details about the credential to the Issuer. It is
+important to note that the timing of such requests could potentially serve as a side-channel.
 
 Verifier/Verifier unlinkablility can be achieved using batch issuance: A batch
 of credentials based on the same claims is issued to the Holder instead of just
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1656,7 +1656,7 @@ data. The original JSON data is then used by the application. See
    * Ensure claims that control validity are checked after decoding payload
    * Update JSON Serialization to remove the kb_jwt member and allow for the disclosures to be conveyed elsewhere
    * Expand the Enveloping SD-JWTs section to also discuss enveloping JSON serialized SD-JWTs
-   * Improve unlinkability considerations, mention that different KB keys must be used.
+   * Improve unlinkability considerations, mention that different KB keys must be used
 
    -05
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1167,7 +1167,7 @@ The following types of unlinkability are considered here:
    learn that they have received presentations from the same user.
  * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
    should not be able to know that a user presented the credential to a certain
-   Verifier that is behaving according to protocol.
+   Verifier that is not behaving maliciously.
  * Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
    credential should not be able to tell that a user presented the credential to
    a certain Verifier, even if the Verifier colludes with the Issuer or becomes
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+
+In both cases, unlinkability is in principle limited to cases where the
+disclosed claims do not contain information that directly or indirectly
+identifies the user. For example, when a tax ID is contained in the disclosed
+claims, the Issuer and Verifier can easily link the user's sessions. However,
+when the user only discloses a birthdate to one Verifier and a postal code to
+another Verifier, the two Verifiers should not be able to tell that they were
+talking to the same user.
+
+With SD-JWT, Issuer/Verifier unlinkablility cannot be achieved, as the issued
```
OR13, line 1099:
```suggestion
Issuer/Verifier unlinkability cannot be achieved with SD-JWT, as the issued
```

OR13 (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
-- Use advanced cryptographic schemes, outside the scope of this specification.
-- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
-
+Unlinkability is a property that prevents an adversary from associating two or
+more sessions of the same user. Two types of unlinkability are often considered:
+
+- Verifier/Verifier unlinkability: Two colluding Verifiers should not be able to
+  tell that they have received presentations from the same user.
+- Issuer/Verifier unlinkability: An Issuer colluding with a Verifier should
+  not able to tell that a user that has received a credential from the Issuer
```
bc-pi, line 1088:
```suggestion
  not be able to tell that a user that has received a credential from the Issuer
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
-- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
-
+Unlinkability is a property that prevents an adversary from associating two or
+more sessions of the same user. Two types of unlinkability are often considered:
+
+- Verifier/Verifier unlinkability: Two colluding Verifiers should not be able to
+  tell that they have received presentations from the same user.
+- Issuer/Verifier unlinkability: An Issuer colluding with a Verifier should
+  not be able to tell that a user that has received a credential from the Issuer
+  has presented it to the Verifier.
```
Sakurann, line 1089:
```suggestion
- Issuer/Verifier unlinkability: An Issuer colluding with a Verifier should
  not be able to tell to which Verifier a user has presented a credential issued from this Issuer.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+In both cases, unlinkability is limited to cases where the
+disclosed claims do not contain information that directly or indirectly
+identifies the user. For example, when a tax ID is contained in the disclosed
+claims, the Issuer and Verifier can easily link the user's sessions. However,
+when the user only discloses a birthdate to one Verifier and a postal code to
+another Verifier, the two Verifiers should not be able to tell that they were
+talking to the same user.
+
+Issuer/Verifier unlinkability cannot be achieved with SD-JWT, as the issued
+SD-JWT is directly forwarded to the Verifier.
```
Sakurann, line 1100: I don't understand this. Three party model in itself allows to realize issuer/verifier unlinkability. Are you trying to say that when the issuer directly forwards to the verifier it is not achieved? But in that case no need for sd-jwt.

```suggestion
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+more sessions of the same user. Two types of unlinkability are often considered:
+
+- Verifier/Verifier unlinkability: Two colluding Verifiers should not be able to
+  tell that they have received presentations from the same user.
+- Issuer/Verifier unlinkability: An Issuer colluding with a Verifier should
+  not be able to tell that a user that has received a credential from the Issuer
+  has presented it to the Verifier.
+
+In both cases, unlinkability is limited to cases where the
+disclosed claims do not contain information that directly or indirectly
```
Sakurann, line 1092: Really? There can be multiple same people with the same birthday. In which case, if the holder discloses only the birthdate to the verifier, the issuer/verifier unlinkability is possible.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+identifies the user. For example, when a tax ID is contained in the disclosed
+claims, the Issuer and Verifier can easily link the user's sessions. However,
+when the user only discloses a birthdate to one Verifier and a postal code to
+another Verifier, the two Verifiers should not be able to tell that they were
+talking to the same user.
+
+Issuer/Verifier unlinkability cannot be achieved with SD-JWT, as the issued
+SD-JWT is directly forwarded to the Verifier.
+
+Verifier/Verifier unlinkablility can be achieved using batch issuance: A batch
```
Sakurann, line 1102:
```suggestion
Verifier/Verifier unlinkablility can be achieved using batch issuance. A batch
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1079,15 +1079,32 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
-or two presentation sessions to the same user on the basis of unique values encoded in the SD-JWT
-(Issuer signature, salts, digests, etc.).
-
-To prevent these types of linkability, various methods, including but not limited to the following ones can be used:
-
-- Use advanced cryptographic schemes, outside the scope of this specification.
-- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
-
+Unlinkability is a property that prevents an adversary from associating two or
+more sessions of the same user. Two types of unlinkability are often considered:
```
Sakurann, line 1083: Is it really sessions..?

Sakurann, line 1083: Is it really sessions..?

Sakurann (changes requested):
I do not agree with some of the text, especially around issuer/verifier ublinkability

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: In the proposed text, the Verifier must collude with multiple Verifiers. I think my original definition is stronger.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: Yes, but in that case, the information was obviously not enough to identify the user.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: The three party model does not realize issuer/verifier unlinkability (per the definition above, it assumes colluding issuers and verifiers). This is one of the major differences between salted-hash based approaches and other approaches using more advanced cryptography.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: I'd like to keep this as-is.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett, line 1083: You're right, 'sessions' is not a concept we used previously. How about this?
```suggestion
Unlinkability is a property that prevents an adversary from associating credential
presentations in certain ways. Two types of unlinkability are often considered:
```

draft-ietf-oauth-selective-disclosure-jwt.md
Denisthemalice: Issuer/Verifier unlinkability: 

The basic property should not include the wording "colluding with a Verifier".

Proposal: An Issuer should not be able to know to which Verifier data derived from a credential issued by the Issuer has been presented to the Verifier.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: `from associating credential presentations of the same user to know about the behavior of that user beyond that user's consent`?
Probably some sentence is needed on why associating is a problem - ie adversary being able to know more about the user than it is supposed to, I think?

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I think your argument is that the example I give above still does not prevent issuer/verifier unlinkability in sd-jwt because the user's public key becomes the correlation factor, which I think needs to be explicitly stated

draft-ietf-oauth-selective-disclosure-jwt.md
OR13: mechanically I think there are 2 things that are trying to fit into a single sentence about "Issuer/Verifier unlinkability" here:

1. phone home / issuer's learning when a specific credential is presented.
2. tracking / fingerprinting of the subject by the issuer.

The first is mitigated partially by heard privacy.
The second requires the issuer not to assign or hide correlation capable identifiers.... and even the signature is such an identifier.

I suggest starting by just writing out the behaviors you want to see, and don't want to see... and then don't give the same name to multiple behaviors.

draft-ietf-oauth-selective-disclosure-jwt.md
OR13: imo the threat is about distinguishability, it has to do with the anonymity set of the data subject... You might find t-closenss / or similar concepts helpful. https://en.wikipedia.org/wiki/T-closeness

Issuer / Verifier: Issuers are distinguishable, Subjects are distinguishable (signature correlation), Holders are not distinguishable (batch issuance).

Verifier / Verifier: Issuers are distinguishable, Subjects are not distinguishable (linked secret), Holders are not distinguishable (zkp).

In all cases, we assume multiple verifiers process presentations and are compromised, and then we evaluate the damage done to data subjects.

OR13: .... what I said above is not right... I think "Issuer / Verifier" unlinkability is a bad name... I think the security property we care about there is timing, metadata analysis.

OR13: Screen Shot 2023-10-30 at 1 53 04 PM

Not sure if this helps, but the reason Issuer / Verifier Unlinkability is not achievable is the signature on the credential... batch issuance does not solve this, it just makes the attacker count unique signatures that are observed in issuers and verifiers.

OR13: I think these labels are possibly the wrong words to be using especially because "verifier / verifier" unlinkability seems stronger than issuer / verifier.... but is actually not.

OR13: Its possible that "single use unlinkability" and "multiple use unlinkability" might be better.

OR13: The privacy pass architecture uses this language:

>  These proofs, or tokens, are
>  private in the sense that a given token cannot be linked to the
>  protocol interaction where that token was initially issued.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: No, what I want to say is: The amount of information that can identify a user varies depending on context knowledge and the set of users that exist. If I'm looking at a club with 12 members all having a different birthdates, the birthdate is sufficient identifying information. If there would be a specific credential for sitting US presidents, unlinkability cannot be achieved at all, even if only the credential type is shared.

So the point is: For batch issuance, even if all keys and so on are fresh for each credential, there is an upper limit on the grade of unlinkability achievable, and it depends on the dataset shared and context knowledge etc.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett, line 1100:
```suggestion
Issuer/Verifier unlinkability cannot be achieved in salted-hash based approaches to credentials, such as SD-JWT, as the issued
credential is directly forwarded to the Verifier.
```

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: The difference is effectively whether the Verifier is cooperating with the attacker or whether it is just following protocol; I'll split this up accordingly.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: @OR13  I'll make a bigger change to this PR hopefully addressing some of your comments, but certainly not everything. I propose that we try to get to an "acceptable" description first, get that merged, and then dig into the details in a separate issue.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,61 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
-or two presentation sessions to the same user on the basis of unique values encoded in the SD-JWT
-(Issuer signature, salts, digests, etc.).
-
-To prevent these types of linkability, various methods, including but not limited to the following ones can be used:
-
-- Use advanced cryptographic schemes, outside the scope of this specification.
-- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
-
+Unlinkability is a property that prevents an adversary from correlating
+credential presentations of the same user (beyond the user's consent).
```
bc-pi, line 1151: This sentence doesn't feel quite right/natural to me. Though I'm not sure this suggestion is particularly great either: 

```suggestion
Unlinkability is a property whereby adversaries are prevented from correlating
credential presentations of the same user (beyond the user's consent).
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,61 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+   originally anonymous information presented to Verifier B, therefore revealing
+   the identities of users of Verifier B.
+
+The following types of unlinkability are considered here:
+
+ * Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
+   learn that they have received presentations from the same user.
+ * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
+   should not be able to tell that a user presented the credential to a certain
+   Verifier (that is behaving according to protocol).
```
peppelinux, line 1170:
```suggestion
 * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
   should not be able to know that a user presented the credential to a certain
   Verifier (that is behaving according to protocol).
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,61 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+ * Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
+   credential should not be able to tell that a user presented the credential to
+   a certain Verifier (even if the Verifier colludes with the Issuer or becomes
+   compromised and leaks stored credentials from presentations).
```
peppelinux, line 1174: @danielfett this is quite similar, or equivalent, to the previous one. If the concrete information is in the brackets probably we might try a rewording without the brackets, WDYT?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,61 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+
+In all cases, unlinkability is limited to cases where the disclosed claims do
+not contain information that directly or indirectly identifies the user. For
```
peppelinux, line 1177:
```suggestion
Unlinkability applies only when the disclosed claims do not hold any information that could directly or indirectly lead to the identification of the user. For
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,61 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+example, when a tax ID is contained in the disclosed claims, the Issuer and
+Verifier can easily link the user's transactions. However, when the user only
+discloses a birthdate to one Verifier and a postal code to another Verifier, the
+two Verifiers should not be able to tell that they were talking to the same
+user.
```
peppelinux, line 1182:
```suggestion
discloses a birthdate to one Verifier and a postal code to another Verifier, the two Verifiers should not be able to determine that they were interacting with the same user.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,61 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be
+achieved in salted-hash based approaches to credentials, such as SD-JWT, as the
+issued credential is directly forwarded to the Verifier, who can forward it to
+the malicious Issuer.
+
+Issuer/Verifier unlinkability with an honest Verifier can be achieved. A
+'phone-home' from the Verifier to the Issuer, for example, for a revocation
+check, SHOULD be avoided, or, if that is not possible, MUST be done in a
+privacy-preserving way where details about the credential are not revealed to
+the Issuer. (Even then, timing of such requests might remain as a side-channel.)
```
peppelinux, line 1193: I didn't fully get this, or probably we might try to simplify this text

peppelinux (changes requested):

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: You're right, let's try to remove the brackets.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,59 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+
+ * Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
+   learn that they have received presentations from the same user.
+ * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
+   should not be able to know that a user presented the credential to a certain
+   Verifier (that is behaving according to protocol).
+ * Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
+   credential should not be able to tell that a user presented the credential to
+   a certain Verifier (even if the Verifier colludes with the Issuer or becomes
+   compromised and leaks stored credentials from presentations).
```
danielfett, line 1174:
```suggestion
 * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
   should not be able to know that a user presented the credential to a certain
   Verifier that is behaving according to protocol.
 * Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
   credential should not be able to tell that a user presented the credential to
   a certain Verifier, even if the Verifier colludes with the Issuer or becomes
   compromised and leaks stored credentials from presentations.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,59 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be
+achieved in salted-hash based approaches to credentials, such as SD-JWT, as the
+issued credential is directly forwarded to the Verifier, who can forward it to
+the malicious Issuer.
+
+Issuer/Verifier unlinkability with an honest Verifier can be achieved. A
+'phone-home' from the Verifier to the Issuer, for example, for a revocation
+check, SHOULD be avoided, or, if that is not possible, MUST be done in a
+privacy-preserving way where details about the credential are not revealed to
+the Issuer. (Even then, timing of such requests might remain as a side-channel.)
```
danielfett, line 1191: How about this?
```suggestion
Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
However, a callback from the Verifier to the Issuer, for example, for a revocation
check, can reveal information about the use of the credential to the Verifier. Therefore,
such callbacks SHOULD be avoided, or, if that is not possible, MUST be done in a
privacy-preserving way where details about the credential are not revealed to
the Issuer. (Even then, timing of such requests might remain as a side-channel.)
```

danielfett: @peppelinux

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: I like the 'limited' wording a bit more, it shows that there is an upper bound.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,59 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
-or two presentation sessions to the same user on the basis of unique values encoded in the SD-JWT
-(Issuer signature, salts, digests, etc.).
-
-To prevent these types of linkability, various methods, including but not limited to the following ones can be used:
-
-- Use advanced cryptographic schemes, outside the scope of this specification.
-- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
-
+Unlinkability is a property whereby adversaries are prevented from correlating
+credential presentations of the same user (beyond the user's consent).
```
peppelinux, line 1151:
```suggestion
Unlinkability is a property that prevents adversaries from correlating credential presentations of the same user beyond the user's consent.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,59 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+   should not be able to know that a user presented the credential to a certain
+   Verifier that is behaving according to protocol.
+ * Issuer/Verifier Unlinkability (Colluding/Compromised Verifier): An Issuer of a
+   credential should not be able to tell that a user presented the credential to
+   a certain Verifier, even if the Verifier colludes with the Issuer or becomes
+   compromised and leaks stored credentials from presentations.
+
+In all cases, unlinkability is limited to cases where the disclosed claims do
+not contain information that directly or indirectly identifies the user. For
+example, when a tax ID is contained in the disclosed claims, the Issuer and
```
peppelinux, line 1178:
```suggestion
example, when a taxpayer identification number is contained in the disclosed claims, the Issuer and
```

draft-ietf-oauth-selective-disclosure-jwt.md
peppelinux: looks good, and I would also contribute with the followings

`However, a callback from the Verifier to the Issuer, for example, for a revocation
check, can reveal information about the use of the credential to the Verifier.` -> `A callback from the Verifier to the Issuer, such as a revocation check, could potentially disclose information about the credential's usage to the Verifier.`

and also

`Therefore,
such callbacks SHOULD be avoided, or, if that is not possible, MUST be done in a
privacy-preserving way where details about the credential are not revealed to
the Issuer. (Even then, timing of such requests might remain as a side-channel.)` -> `Therefore, such callbacks SHOULD be avoided. If they are unavoidable, they MUST be executed in a manner that preserves privacy and does not disclose details about the credential to the Issuer. However, it's important to note that the timing of such requests could potentially serve as a side-channel.`

peppelinux (approved):
@danielfett I share my last comments that I'm sure that you'll consider 

approved, thank you!

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett:
```suggestion
Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
disclose information about the credential's usage to the Verifier. Therefore, such callbacks
SHOULD be avoided. Where they are unavoidable, they MUST be executed in a manner that
preserves privacy and does not disclose details about the credential to the Issuer. It is
important to note that the timing of such requests could potentially serve as a side-channel.
```

draft-ietf-oauth-selective-disclosure-jwt.md
peppelinux: likewise, dont worry and go ahead while you can :-)

draft-ietf-oauth-selective-disclosure-jwt.md
peppelinux: hey Dan, I don't love "to tell" and the concept of the speacking machines :-)

WDYT about this

````
Issuer/Verifier unlinkability: If an Issuer and a Verifier were to collude, they should not be able to determine that a user has presented a credential from the Issuer to that Verifier.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,60 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+In all cases, unlinkability is limited to cases where the disclosed claims do
+not contain information that directly or indirectly identifies the user. For
+example, when a taxpayer identification number is contained in the disclosed claims, the Issuer and
+Verifier can easily link the user's transactions. However, when the user only
+discloses a birthdate to one Verifier and a postal code to another Verifier, the two Verifiers should not be able to determine that they were interacting with the same user.
+
+Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be
+achieved in salted-hash based approaches to credentials, such as SD-JWT, as the
+issued credential is directly forwarded to the Verifier, who can forward it to
+the malicious Issuer.
```
Sakurann, line 1185:
```suggestion
Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be
achieved in salted-hash based selective disclosure approaches to credentials, such as SD-JWT, as the
issued credential with the Issuer's signature is directly forwarded to the Verifier, who can forward it to
the Issuer, if the Issuer is malicious.
```

forwarding to the malicious Issuer part felt a little abrupt on the first read.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,60 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
-
-To prevent these types of linkability, various methods, including but not limited to the following ones can be used:
-
-- Use advanced cryptographic schemes, outside the scope of this specification.
-- Issue a batch of SD-JWTs to the Holder to enable the Holder to use a unique SD-JWT per Verifier. This only helps with Verifier/Verifier unlinkability.
-
+Unlinkability is a property whereby adversaries are prevented from correlating
+credential presentations of the same user beyond the user's consent.
+Otherwise, an adversary might be able to learn more about the user than the user
+intended to disclose, for example:
```
Sakurann, line 1153:
```suggestion
Without unlinkability, an adversary might be able to learn more about the user than the user
intended to disclose, for example:
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,60 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+   originally anonymous information presented to Verifier B, therefore revealing
+   the identities of users of Verifier B.
+
+The following types of unlinkability are considered here:
+
+ * Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
+   learn that they have received presentations from the same user.
+ * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
+   should not be able to know that a user presented the credential to a certain
+   Verifier that is behaving according to protocol.
```
Sakurann, line 1170:
```suggestion
 * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
   should not be able to know that a user presented the credential to a certain
   Verifier that is behaving according to a protocol.
```
what is meant by protocol, here..?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,60 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+achieved in salted-hash based approaches to credentials, such as SD-JWT, as the
+issued credential is directly forwarded to the Verifier, who can forward it to
+the malicious Issuer.
+
+Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
+However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
+disclose information about the credential's usage to the Verifier. Therefore, such callbacks
+SHOULD be avoided. Where they are unavoidable, they MUST be executed in a manner that
+preserves privacy and does not disclose details about the credential to the Issuer. It is
+important to note that the timing of such requests could potentially serve as a side-channel.
```
Sakurann, line 1192:
```suggestion
Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
disclose information about the credential's usage to the Issuer. Therefore, such callbacks
SHOULD be avoided. Where they are unavoidable, they MUST be executed in a manner that
preserves privacy and does not disclose details about the credential to the Issuer. It is
important to note that the timing of such requests could potentially serve as a side-channel.
```

Sakurann (commented):
thank you! feels much clearer. 
could we include somewhere that issuer's signature can be used as a correlating factor?

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett:
```suggestion
 * Issuer/Verifier Unlinkability (Honest Verifier): An Issuer of a credential
   should not be able to know that a user presented the credential to a certain
   Verifier that is not behaving maliciously.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,60 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+ * Issuers might want to track where users present their credentials to enable
+   surveillance.
+ * After a data breach at multiple Verifiers, publicly available information
+   might allow linking identifiable information presented to Verifier A with
+   originally anonymous information presented to Verifier B, therefore revealing
+   the identities of users of Verifier B.
+
+The following types of unlinkability are considered here:
+
+ * Verifier/Verifier Unlinkability: Two colluding Verifiers should not be able to
```
Sakurann, line 1166: during OID4VCI privacy considerations discussion @jogu talked about the same verifier correlating the user across multiple presentations.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,60 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+
+In all cases, unlinkability is limited to cases where the disclosed claims do
+not contain information that directly or indirectly identifies the user. For
+example, when a taxpayer identification number is contained in the disclosed claims, the Issuer and
+Verifier can easily link the user's transactions. However, when the user only
+discloses a birthdate to one Verifier and a postal code to another Verifier, the two Verifiers should not be able to determine that they were interacting with the same user.
+
+Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be
+achieved in salted-hash based selective disclosure approaches, such as SD-JWT, as the
+issued credential with the Issuer's signature is directly forwarded to the Verifier, who can forward it to
```
Sakurann, line 1184:
```suggestion
issued credential with the Issuer's signature is directly presented to the Verifier, who can forward it to
```
the first `directly presented/forwarded` is upon user consent from the wallet to the verifier, right? if so, presented feels more natural than forwarding..?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,60 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+the Issuer.
+
+Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
+However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
+disclose information about the credential's usage to the Issuer. Therefore, such callbacks
+SHOULD be avoided. Where they are unavoidable, they MUST be executed in a manner that
```
Sakurann, line 1190: I am not sure I am ok with `SHOULD be avoided`, given that there are ways to abstract from the issuer who is calling its status endpoint for example.

```suggestion
disclose information about the credential's usage to the Issuer.
Where such callbacks are necessary, they MUST be executed in a manner that
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1147,15 +1147,62 @@ Decoy digests increase the size of the SD-JWT. The number of decoy digests (or w
...
+the Issuer.
+
+Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved.
+However, a callback from the Verifier to the Issuer, such as a revocation check, could potentially
+disclose information about the credential's usage to the Issuer.
+Where such callbacks are necessary, they MUST be executed in a manner that
+preserves privacy and does not disclose details about the credential to the Issuer. It is
+important to note that the timing of such requests could potentially serve as a side-channel.
+
+Verifier/Verifier unlinkablility can be achieved using batch issuance: A batch
```
bc-pi, line 1196:
```suggestion
Verifier/Verifier unlinkablility and presentation unlinkablility can be achieved using batch issuance: A batch
```

Now that https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/354/commits/a1cbe276fcd1b5d85ac00a8985e3fd0fb7039f13 introduced presentation unlinkability and the flow of this whole section (where the unlinkablility types are first described and then discussed) it seems like it warrants a mention here. WDYT?

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: This is solved now.

Denisthemalice:
Instead of one section (12.4) about "Unlinkability", there should be two sections:

"Unlinkability between Verifiers" means that two Verifiers when colluding, should not be able to know whether two digital presentations are presented by the same user.

"Untrackability by an issuer" means that an Issuer should not be able to know to which Verifier a digital presentation will be or has been presented by a user.

Note:  In the second case, there is no need to have a collusion between an Issuer and a Verifier.

Denisthemalice:
Instead of one section about "Unlinkability", there should be two sections:

"**Unlinkability between Verifiers**" means that two Verifiers when colluding, should not be able to know whether two digital presentations are presented by the same user.

"**Untrackability by an issuer**" means that an Issuer should not be able to know to which Verifier a digital presentation will be or has been presented by a user.

Note:  In the second case, there is no need to have a collusion between an Issuer and a Verifier.

bc-pi (approved):

Sakurann (approved):

danielfett:
For the sake of having a starting point, the authors decided to merge this pull request to have a starting point in the document. The current text seems to be a rough consensus. We are open to modifying the text later on, as already mentioned above.

PR #361, opened 2023-10-27
Reset branch for example generation

danielfett:


bc-pi (approved):
see that https://github.com/openwallet-foundation-labs/sd-jwt-python/pull/8 was merged so this seems good

Sakurann (approved):

PR #362, opened 2023-10-31
Update change controller for the Structured Syntax Suffix registration

bc-pi:
Update change controller for the Structured Syntax Suffix registration from IESG to IETF per IANA suggestion

### DIFF
--- a/.github/workflows/ghpages.yml
+++ b/.github/workflows/ghpages.yml
@@ -51,7 +51,7 @@ jobs:
         restore-keys: i-d-
 
     - name: "Build Drafts"
-      uses: martinthomson/i-d-template@3e5cfbafc0037fee41e7fe4e6567b14b3c1be199
+      uses: martinthomson/i-d-template@v1
       with:
         token: ${{ github.token }}
 
--- a/.github/workflows/ghpages.yml
+++ b/.github/workflows/ghpages.yml
@@ -51,7 +51,7 @@ jobs:
         restore-keys: i-d-
 
     - name: "Build Drafts"
-      uses: martinthomson/i-d-template@v1
+      uses: martinthomson/i-d-template@3e5cfbafc0037fee41e7fe4e6567b14b3c1be199
       with:
         token: ${{ github.token }}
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -8,4 +8,5 @@ ifneq (,$(shell grep "path *= *$(LIBDIR)" .gitmodules 2>/dev/null))
 else
 	git clone -q --depth 10 $(CLONE_ARGS) \
 	    -b main https://github.com/martinthomson/i-d-template $(LIBDIR)
+	(cd $(LIBDIR) && git checkout 3e5cfbafc0037fee41e7fe4e6567b14b3c1be199) # Temporarily pin to a specific commit right before it upgraded mmark from 2.2.10 to 2.2.40 as the latter is producing XML from the current markdown that xml2rfc rejects
```
bc-pi, line 11: Well, that didn't work. Apparently I don't understand the github workflow/action build process well enough.

c2bo: The dependencies (like mmark) seem to be installed in the docker container of the github action -> You would need to modify the action version to switch to an older mmark version from what I understand?

bc-pi: Maybe? I'm honestly not sure how to do that or even which action or how/if its versioned. 

https://github.com/martinthomson/i-d-template/pull/412/files is the change to action/Dockerfile in the martinthomson/i-d-template FWIW. 

I was just trying for quick workaround here though. And chasing down the main issue later. But maybe a quick workaround isn't doable.

bc-pi: I clearly have no idea what I'm doing ... :(

.github/workflows/ghpages.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         restore-keys: i-d-
 
     - name: "Build Drafts"
-      uses: martinthomson/i-d-template@v1
+      uses: martinthomson/i-d-template@3e5cfbafc0037fee41e7fe4e6567b14b3c1be199
```
bc-pi, line 54: that doesn't work as i'd naively hoped either

danielfett (approved):

PR #380, opened 2023-11-15
Strengthen security considerations around validity claims not being selectively disclosable

### LINKED ISSUE #367: selectively disclosable
from this thread https://mailarchive.ietf.org/arch/msg/oauth/5A1kXMNiIm2DBEcXCadCwUpMhgI/ Neil Madden has suggested some text (below) for the [Selectively-Disclosable Validity Claims](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-06.html#name-selectively-disclosable-val) section.  I think we can replace and/or augment and/or merge what's there with this suggestion and arrive at something reasonable that's a bit stronger about not making claims that control the validity of the SD-JWT selectively disclosable. 

> 
> An Issuer MUST NOT allow any security-critical claim to be selectively disclosable. The exact list of “security-critical” claims will depend on the application, and SHOULD be listed by any application-specific profile of SD-JWT. The following is a list of standard claim names that SHOULD be considered as security-critical by any SD-JWT Issuer:
> 
>  * “iss” (Issuer)
>  * “aud” (Audience), although issuers may want to allow individual entries in the array to be selectively-disclosable
>  * “exp” (Expiration Time)
>  * “nbf” (Not Before)
>  * “iat” (Issued At)
>  * “jti” (JWT ID)
> 
> In addition, the “cnf” (Confirmation Key) claim MUST NOT be selectively disclosable.

bc-pi:
Strengthen security considerations around claims controlling the validity of the SD-JWT not being selectively disclosable

to fix #367

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -960,21 +960,37 @@ the SD-JWT.
 
 ## Selectively-Disclosable Validity Claims {#sd-validity-claims}
 
-Claims controlling the validity of the SD-JWT, such as `nbf`, `iat`, and `exp`,
-are usually included in plaintext in the SD-JWT payload, but MAY be
-selectively disclosable instead. In this case, however, it is up to the Holder
-to release the claims to the Verifier. A malicious Holder may try to hide, for
+An Issuer MUST NOT allow any security-critical claim to be selectively disclosable.
+The exact list of such claims will depend on the application,
+and SHOULD be listed by any application-specific profile of SD-JWT.
+The following is a list of standard claim names that SHOULD be considered as
+security-critical by any SD-JWT Issuer:
+
+* `iss` (Issuer)
+* `aud` (Audience), although issuers may want to allow individual entries in the array to be selectively disclosable
+* `exp` (Expiration Time)
+* `nbf` (Not Before)
+* `iat` (Issued At)
+* `jti` (JWT ID)
+
+In addition, the `cnf` (Confirmation Key) claim MUST NOT be selectively disclosable.
+
+Consequently, claims controlling the validity of the SD-JWT
+will typically be included in plaintext in the SD-JWT payload. Verifiers, however, cannot
+reliably depend on that and need to operate as though security-critical claims might be
+selectively disclosable. In such a case, the onus is on the Holder
+to release the relevant claims to the Verifier. And a malicious Holder may try to hide, for
 example, an expiration time (`exp`) in order to get a Verifier that "fails open"
 to accept an expired SD-JWT.
 
 Verifiers therefore MUST ensure that all claims they deem necessary for checking
-the validity of the SD-JWT are present (or disclosed, respectively) before
-checking the validity and accepting the SD-JWT. This is implemented in the last
+the validity of an SD-JWT in the given context are present (or disclosed, respectively) during
+validation of the SD-JWT. This is implemented in the last
 step of the verification defined in (#sd_jwt_verification).
 
 The precise set of required validity claims will typically be defined by
-ecosystem rules or the credential format and MAY include claims other than
-`nbf`, `iat`, and `exp`.
+ecosystem rules, application-specific profile, or the credential format and MAY include claims other than
+those listed herein.
 
 ## Issuer Signature Key Distribution and Rotation {#issuer_signature_key_distribution}
 
@@ -1143,6 +1159,7 @@ Matthew Miller,
 Mike Jones,
 Mike Prorock,
 Nat Sakimura,
+Neil Madden,
 Oliver Terbu,
 Orie Steele,
 Paul Bastian,
@@ -1661,6 +1678,7 @@ data. The original JSON data is then used by the application. See
    -07
 
    * Update change controller for the Structured Syntax Suffix registration from IESG to IETF per IANA suggestion
+   * Strengthen security considerations around claims controlling the validity of the SD-JWT not being selectively disclosable
 
    -06
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -975,7 +975,8 @@ the SD-JWT.
 
 ## Selectively-Disclosable Validity Claims {#sd-validity-claims}
 
-An Issuer MUST NOT allow any security-critical content to be selectively disclosable.
+An Issuer MUST NOT allow any content to be selectively disclosable that is critical for evaluating the
+SD-JWT's authenticity or validity.
 The exact list of such content will depend on the application
 and SHOULD be listed by any application-specific profile of SD-JWT.
 The following is a list of typical JWT claim names that SHOULD be considered as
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -960,21 +960,37 @@ the SD-JWT.
...
+security-critical by any SD-JWT Issuer:
+
+* `iss` (Issuer)
+* `aud` (Audience), although issuers may want to allow individual entries in the array to be selectively disclosable
+* `exp` (Expiration Time)
+* `nbf` (Not Before)
+* `iat` (Issued At)
+* `jti` (JWT ID)
+
+In addition, the `cnf` (Confirmation Key) claim MUST NOT be selectively disclosable.
```
Sakurann, line 976: why? 
Issuer might be using claims other than `cnf` to do key binding?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -960,21 +960,37 @@ the SD-JWT.
...
+and SHOULD be listed by any application-specific profile of SD-JWT.
+The following is a list of standard claim names that SHOULD be considered as
+security-critical by any SD-JWT Issuer:
+
+* `iss` (Issuer)
+* `aud` (Audience), although issuers may want to allow individual entries in the array to be selectively disclosable
+* `exp` (Expiration Time)
+* `nbf` (Not Before)
+* `iat` (Issued At)
+* `jti` (JWT ID)
```
Sakurann, line 974: I am not sure about the format of this... SD-JWT is payload agnostic, so there is no guarantee the payload uses JWT claims (though it is of course a convention) - for example, W3C VCDM might use their own claims (`issuer`, `validFrom`, etc.)

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -960,21 +960,37 @@ the SD-JWT.
 
 ## Selectively-Disclosable Validity Claims {#sd-validity-claims}
 
-Claims controlling the validity of the SD-JWT, such as `nbf`, `iat`, and `exp`,
-are usually included in plaintext in the SD-JWT payload, but MAY be
-selectively disclosable instead. In this case, however, it is up to the Holder
-to release the claims to the Verifier. A malicious Holder may try to hide, for
+An Issuer MUST NOT allow any security-critical claim to be selectively disclosable.
+The exact list of such claims will depend on the application,
+and SHOULD be listed by any application-specific profile of SD-JWT.
```
danielfett, line 965:
```suggestion
The exact list of such claims will depend on the application
and SHOULD be listed by any application-specific profile of SD-JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: Maybe an explicit reference to this being the default JWT claims would make this clear.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -960,21 +960,37 @@ the SD-JWT.
...
+The following is a list of standard claim names that SHOULD be considered as
+security-critical by any SD-JWT Issuer:
+
+* `iss` (Issuer)
+* `aud` (Audience), although issuers may want to allow individual entries in the array to be selectively disclosable
```
danielfett, line 970:
```suggestion
* `aud` (Audience), although issuers may want to allow individual entries in the array to be selectively disclosable
```
Uppercase MAY? Not sure...

danielfett (commented):
Overall this looks good to me, thanks for preparing. I think the list of claims should be introduced as "typical" or we should not that it is relevant specifically for JWT-based payloads.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -975,21 +975,34 @@ the SD-JWT.
 
 ## Selectively-Disclosable Validity Claims {#sd-validity-claims}
 
-Claims controlling the validity of the SD-JWT, such as `nbf`, `iat`, and `exp`,
-are usually included in plaintext in the SD-JWT payload, but MAY be
-selectively disclosable instead. In this case, however, it is up to the Holder
-to release the claims to the Verifier. A malicious Holder may try to hide, for
+An Issuer MUST NOT allow any security-critical content to be selectively disclosable.
```
danielfett, line 978: Maybe we should be more specific here:
```suggestion
An Issuer MUST NOT allow any content to be selectively disclosable that is critical for evaluating the
SD-JWT's authenticity or validity.
```
(Not sure if `authenticity or validity` captures all things we might want to say here.)

bc-pi: I think it probably captures enough.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -994,21 +994,35 @@ the SD-JWT.
...
 
-Claims controlling the validity of the SD-JWT, such as `nbf`, `iat`, and `exp`,
-are usually included in plaintext in the SD-JWT payload, but MAY be
-selectively disclosable instead. In this case, however, it is up to the Holder
-to release the claims to the Verifier. A malicious Holder may try to hide, for
+An Issuer MUST NOT allow any content to be selectively disclosable that is critical for evaluating the
+SD-JWT's authenticity or validity.
+The exact list of such content will depend on the application
+and SHOULD be listed by any application-specific profile of SD-JWT.
+The following is a list of typical JWT claim names that SHOULD be considered as
```
Sakurann, line 1001:
```suggestion
The following is a list of registered JWT claim names that SHOULD be considered as
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -994,21 +994,35 @@ the SD-JWT.
...
+* `nbf` (Not Before)
+* `cnf` (Confirmation Key)
+
+Consequently, claims controlling the validity of the SD-JWT
+will typically be included in plaintext in the SD-JWT payload. Verifiers, however, cannot
+reliably depend on that and need to operate as though security-critical claims might be
+selectively disclosable. In such a case, the onus is on the Holder
+to release the relevant claims to the Verifier. And a malicious Holder may try to hide, for
 example, an expiration time (`exp`) in order to get a Verifier that "fails open"
 to accept an expired SD-JWT.
```
Sakurann, line 1016: sorry, I don't fully understand the purpose of this paragraph.. 
if issuer will typically put in the plaintext, why verifier cannot rely on that?
If there is a possibility of a malicious holder, shouldn't the sentence be phrased more like a security consideration for the verifiers not to "fail open"..?

```suggestion
Consequently, Issuer will typically include claims controlling the validity of the SD-JWT
in plaintext in the SD-JWT payload. Verifiers, however, cannot
reliably depend on that and need to operate as though security-critical claims might be
selectively disclosable. In such a case, the onus is on the Holder
to release the relevant claims to the Verifier. And a malicious Holder may try to hide, for
example, an expiration time (`exp`) in order to get a Verifier that "fails open"
to accept an expired SD-JWT.
```

bc-pi: Your suggested text reads much better and I think we can do without the malicious holder example now.

Sakurann (approved):

danielfett (approved):

Issue #369, opened 2023-11-07
Allow HMAC-based Issuer-signed JWTs and KB-JWTs to support different aspects of repudiation


awoie:
Currently the spec says:
> It MUST NOT use none or an identifier for a symmetric algorithm (MAC).

To support KB-JWTs using HS256, for example after using DH key agreement, we should allow symmetric algorithms.

tlodderstedt:
I support this issue. There are scenarios where sd-jwts may be protected with symmetric crypto. I'm wondering why symmetric algorithm (MAC) are ruled out so far.

awoie:
The section I quoted applies to Issuer-signed JWTs, closing this issue.

awoie:
After some internal discussion, I guess it is worthwhile to confirm with the authors that HMAC for KB-JWT is supported @bc-pi @danielfett ?

awoie:
I quoted the wrong section above. However, in the KB-JWT section there is a similar requirement on `alg` of the KB-JWT:
> alg: REQUIRED. A digital signature algorithm identifier such as per IANA "JSON Web Signature and Encryption Algorithms" registry. MUST NOT be none or an identifier for a symmetric algorithm (MAC).

By having this requirement it means the SD-JWT spec does not allow HS256 and therefore does not allow approaches where a HS256 is agreed via DH. Note, the asymmetric DH key could still be endorsed by the `cnf` claim

tlodderstedt:
I think it should be allowed for both, kb jwt and sd jwt. Or let me put it the other way around. Why is it prohibited now?

paulbastian:
The text is only about SD-JWT right now. I actually want HMAC for SD-JWT

awoie:
> The text is only about SD-JWT right now. I actually want HMAC for SD-JWT

Currently, the spec does not allow HMAC for Issuer-signed JWT and it does not allow HMAC for KB-JWTs. To support repudiation KB-JWTs should allow `alg` values involving symmetric keys such as HS256 (see my comment above).

Sakurann:
I don't support adding an option to HMAC KB JWT. Issuer signs over the key used to HMAC KB JWT. Anyone who knows the nonce of the transaction can generate that HMAC KB JWT. It does not proof anything. It only makes sense if it is issuance during presentation where wallet is asking issuer to issue over wallet's key that wallet generated for that verifier.

awoie:
[...]

PR #400, opened 2024-02-13
Do not disallow HMAC any longer.

danielfett:
This addresses Issue #369. The rest of the text still refers to asymmetric signatures, but technically, an HMAC is now allowed.


bc-pi (approved):

paulbastian (commented):
Just noting that an MAC issuer key is usually not referred to as a private key. Maybe put private in brackets, but I'm unsure.

bc-pi:
> Just noting that an MAC issuer key is usually not referred to as a private key. Maybe put private in brackets, but I'm unsure.

The scope of this PR was intentionally to only remove the explicit prohibition on MAC.

paulbastian (approved):

Sakurann (approved):

Issue #383, opened 2023-11-22
Remove mention of unspecified key binding methods


bifurcation:
The algorithm in [Section 8.3](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-06.html#section-8.3) includes the following branch:

> If Key Binding is provided by means not defined in this specification, verify the Key Binding according to the method used.

We should remove that branch.  If an implementation is doing something outside this specification, their behavior can deviate arbitrarily from what the specification says.  It's not the job of this specification to cover everything that a verifier might possibly do.

bc-pi:
Indeed it's not the job of this specification to cover everything that a verifier might possibly do. But verifying key binding is important and the specification elsewhere discusses the possibility that the proof of possession could be accomplished by ways other than the KB JWT. As such, we felt that it deserved treatment in the validation steps.

bifurcation:
On the one hand, KB is important.  On the other hand, KB is optional.  🤔 

What other discussion do you have in mind?  I'm not finding it on a quick search through the document.

bc-pi:
[Alternatives to a Key Binding JWT](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-06#name-alternatives-to-a-key-bindi) for example

bifurcation:
Yeah, we should delete that and Section 10 as well.  They don't actually define anything, and they are harmful because they encourage divergent, non-interoperable implementations.  If there are use cases that KB-JWT doesn't cover and needs to, we should accommodate them.  If we don't need to accommodate them, we shouldn't.

Sakurann:
discussed during the editor's call - agreed to remove specific references that allow additional key binding mechanisms, to encourage interoperability using the mechanisms defined in the spec and because even without the text those additional mechanisms are not precluded as spec makes it clear KB JWT is optional.

@danielfett to open a separate issue on updating section 9 to reflect most recent implementation experience.

bc-pi:
PR #404 removes mention of unspecified key binding methods and the Enveloping SD-JWTs section

bc-pi:
> discussed during the editor's call - agreed to remove specific references that allow additional key binding mechanisms, to encourage interoperability using the mechanisms defined in the spec and because even without the text those additional mechanisms are not precluded as spec makes it clear KB JWT is optional.

[...]

PR #404, opened 2024-02-20
Remove mention of unspecified key binding methods

bc-pi:
Remove mention of unspecified key binding methods and the Enveloping SD-JWTs section (issue #383)

Editor's copy preview of this PR: https://drafts.oauth.net/oauth-selective-disclosure-jwt/remove-mention-of-unspecified-key-binding-methods/draft-ietf-oauth-selective-disclosure-jwt.html

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -780,7 +780,7 @@ application or transport protocol. However, the details of such approaches fall 
 specification.
 
 Verification of the JWS JSON serialized SD-JWT follows the same rules defined in (#verification),
-except that the SD-JWT does not need to be split into component parts, the disclosures
+except that the SD-JWT does not need to be split into component parts and the disclosures
 can be found in the respective member of the JSON object (or elsewhere).
 
 Using a payload similar to that from [Example 1](#example-1), the following is a non-normative example of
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -791,8 +781,7 @@ specification.
 
 Verification of the JWS JSON serialized SD-JWT follows the same rules defined in (#verification),
 except that the SD-JWT does not need to be split into component parts, the disclosures
```
danielfett, line 783:
```suggestion
except that the SD-JWT does not need to be split into component parts and the disclosures
```

danielfett (approved):
Thank you, approved with one nit.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -756,17 +748,15 @@ To this end, Verifiers MUST follow the following steps (or equivalent):
    details.
 2. Process the SD-JWT as defined in (#sd_jwt_verification).
 3. If Key Binding is required:
```
Sakurann, line 750:
```suggestion
3. If Key Binding JWT is required:
```

if Key Binding is required, but can be provided using means other than KB JWT, the current 3.1 is too strict

Sakurann (approved):
noting that this change solidifies that to be compliant to the sd-jwt spec, one must use kb jwt for key binding, more than merely removing a note that key binding can be provided using means other than kb jwt.

PR #408, opened 2024-02-27
Fix the PID example

danielfett:
Remove `vct#integrity` that is not yet defined and remove the disclosure of the age which for whatever reason didn't work.

### DIFF
--- a/examples/arf-pid/specification.yml
+++ b/examples/arf-pid/specification.yml
@@ -1,6 +1,6 @@
 user_claims:
   vct: https://bmi.bund.example/credential/pid/1.0
-  vct#integrity: sha256-jo8433ot48utul8ura33
+  #vct#integrity: sha256-jo8433ot48utul8ura33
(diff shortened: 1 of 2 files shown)


examples/arf-pid/specification.yml
```diff
@@ -37,8 +37,8 @@ user_claims:
 holder_disclosed_claims:
   nationalities:
     - true
-  age_equal_or_over:
-    '18': true
+  #age_equal_or_over:
```
bc-pi, line 40: We should probably chase this down sometime sooner rather than later (some issue with the yaml or a bug in the library maybe?).

bc-pi (approved):
Looks okay to address the immediate issues now.

examples/arf-pid/specification.yml
bc-pi: created "disclosing age_equal_or_over not working" #409

PR #410, opened 2024-02-28
more consistent treatment of a Disclosure vs its contents

### LINKED ISSUE #358: Confusing terminology
Hi,

Disclosure is used at least two ways in the document.

- to refer to the array containing the salt, (maybe name), and value.
- to refer to the base64url encoded string 

It's easy to understand how this could have happened, as both contain logically the same information. I propose explicitly calling these:

- **disclosure array**, and 
- **disclosure presentation**

I also propose calling the hash/digest of the disclosure the **disclosure digest** in full in every use. The only time we would use "disclosure" without one of these modifiers is talking about the process of disclosing in general.

As for the plaintext of the claims that can be disclosed, I propose **blinded claim**.

So the first paragraph of Section 4.1 would read:

An SD-JWT consists of:

- a signed JSON document, 
- a list of _disclosure presentations_ (each of which reveals a _blinded claim_), and 
- optionally an _issuer key binding_ (discussed in Section 4.3). 

The signed JSON document (the JWT body) contains plaintext claims; and _disclosure digests_, each of which either refers to a specific _blinded claim_ or is a _decoy_. The holder can include zero or more disclosure presentations without breaking the signature of the JWT body. The contents of the blinded claims cannot be modified, because the corresponding digest would no longer match any digest in the JWT body. Blinded claims can be individual object properties (key-value pairs) or array elements.

bc-pi:
Editorial updatess for more consistent treatment of a Disclosure vs the contents of a Disclosure (to fix #358 Confusing terminology)

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -149,7 +149,7 @@ Selectively Disclosable JWT (SD-JWT):
  that supports selective disclosure as defined in this document. It can contain both regular claims and digests of selectively-disclosable claims.
 
 Disclosure:
-:  A combination of a salt, a cleartext claim name (present when the claim is a name-value pair and absent when the claim is an array element), and a cleartext claim value, all of which are used to calculate a digest for the respective claim.
+:  A JSON array containing a combination of a salt, a cleartext claim name (present when the claim is a name-value pair and absent when the claim is an array element), and a cleartext claim value, which is base64url-encoded and used to calculate a digest for the respective claim. The term Disclosure refers to the whole base64url-encoded string.
 
 Key Binding:
 :  Ability of the Holder to prove legitimate possession of an SD-JWT by proving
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -430,9 +431,9 @@ The resulting Disclosure would be: `WyJsa2x4RjVqTVlsR1RQVW92TU5JdkNBIiwgIkZSIl0`
 
 ### Hashing Disclosures {#hashing_disclosures}
 
-For embedding the Disclosures in the SD-JWT, the Disclosures are hashed using the hash algorithm specified in the `_sd_alg` claim described in (#hash_function_claim). The resulting digest is then included in the SD-JWT payload instead of the original claim value, as described next.
+For embedding references to the Disclosures in the SD-JWT, each Disclosure is hashed using the hash algorithm specified in the `_sd_alg` claim described in (#hash_function_claim). The resulting digest is then included in the SD-JWT payload instead of the original claim value, as described next.
 
-The digest MUST be taken over the US-ASCII bytes of the base64url-encoded Disclosure. This follows the convention in JWS [@!RFC7515] and JWE [@RFC7516]. The bytes of the digest MUST then be base64url-encoded.
+The digest MUST be taken over the US-ASCII bytes of the base64url-encoded value that is the Disclosure. This follows the convention in JWS [@!RFC7515] and JWE [@RFC7516]. The bytes of the digest MUST then be base64url-encoded.
```
Sakurann, line 436:
```suggestion
The digest MUST be taken over the Disclosure, which is US-ASCII bytes of the base64url-encoded value. This follows the convention in JWS [@!RFC7515] and JWE [@RFC7516]. The bytes of the digest MUST then be base64url-encoded.
```

Sakurann (approved):
made one suggestion, but overall, looks good to me! thank you!

danielfett (approved):
Looks good to me, thank you!

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: "i.e." would work slightly better than the "which is" in this case, but I'm also fine with the original wording.

bc-pi: ok w/ not changing

Issue #409, opened 2024-02-27
disclosing age_equal_or_over not working


bc-pi:
> We should probably chase this down sometime sooner rather than later 
> (some issue with the yaml or a bug in the library maybe?).

_Originally posted by @bc-pi in https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/408#discussion_r1505105674_

PR #411, opened 2024-02-29
Fix PID example's age disclosure

bc-pi:
Fix the disclosure of the age >= 18 in the PID example (something with type conversion and quoting in the yaml) and remove the not yet defined vct#integrity

fixes issue #409

alternative to PR #408 

https://drafts.oauth.net/oauth-selective-disclosure-jwt/fix-age-in-pid-example/draft-ietf-oauth-selective-disclosure-jwt.html#appendix-A.3-57 does now have ~ `"age_equal_or_over": {"18": true} ` show up


Sakurann (approved):

danielfett:
I wouldn't have expected that this fixes anything, but TIL I guess?

Issue #403, opened 2024-02-20
Redesign JWS JSON Serialization


danielfett:
Through discussions in the EUDI context, we found that the current JSON Serialization design is not well-suited for existing implementations. We now also better understand the original use cases that lead us to include this section into the spec in the first place and we agreed that we should update it as follows:

* The disclosures should go into the unprotected header (to be defined: What happens when there are multiple signatures and therefore multiple unprotected headers with disclosures?)
* There should be a way to put the key binding JWT into the unprotected header (to be defined: What happens when there are multiple signatures and therefore multiple places where a KB JWT could go?). The way the hash is built could remain unchanged, but the order of disclosures needs to be clearly defined if they can show up in multiple headers. This would also solve #392.

The new flattened format could be:
```json
{
  "payload":"eyJfc2QiOiBbIjR...Aic2hhLTI1NiJ9",
  "protected":"eyJhbGciOiAiRVMyNTYiLCAidHlwIjogImV4YW1wbGUrc2Qtand0In0",
  "signature":"oL5lornrs6CFBMu...4HQby6OaaLqAw",
  "header":{
    "disclosures":[
      "WyI2SWo3dE0tYTVpVlBH...W1lIiwgIkRvZSJd",
      "WyJBSngtMDk1VlBycFR...ImFkZHJlc3MiLCB7InN0cmVldF9",
      "WyJlbHVWNU9nM2dTT...VuX25hbWUiLCAiSm9obiJd"
    ],
    "kb_jwt":"eyJhbGciOiAiRVMyNTYiLCAidHlwIjogI...bGUrc2Qtand0In0"
  }
}
```

And the new general format could be:
```json
{
  "payload":"eyJfc2QiOiBbIjR...Aic2hhLTI1NiJ9",
  "signatures":[
    {
      "protected":"eyJhbGciOiAiRVMyNTYiLCAidHlwIjogImV4YW1wbGUrc2Qtand0In0",
      "signature":"oL5lornrs6CFBMu...4HQby6OaaLqAw",
      "header":{
        "disclosures":[
          "WyI2SWo3dE0tYTVpVlBH...W1lIiwgIkRvZSJd",
          "WyJBSngtMDk1VlBycFR...ImFkZHJlc3MiLCB7InN0cmVldF9",
          "WyJlbHVWNU9nM2dTT...VuX25hbWUiLCAiSm9obiJd"
        ],
[...]

Issue #392, opened 2023-12-04
JSON serialization doesn't cover key binding


bifurcation:
As in #383, [9. JWS JSON Serialization](https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#name-jws-json-serialization) invokes specification by magic:

> Key Binding (if applicable) will be provided by means not specifically defined in this specification

If we are defining a format for SD-JWT, that format needs to include all of the parts of the SD-JWT.  Unless #374 gets fixed, that means the format needs an optional key binding field.

But I think that this issue is actually a symptom #374.  The document doesn't define a format for "SD-JWT with KB" because it is actually substantively different thing from "SD-JWT without KB".  Putting it in JSON just highlights the difference more clearly than `~`-delimited text.

The current syntax seems fine to me for "SD-JWT without KB".  For "SD-JWT with KB", I would think you would want something like:

```
{
  "sd_jwt": { /* format currently specified, or compact */ },
  "key_binding: { /* JSON-formatted JWS of KB-JWT */ }
}
```

Here the `sd_jwt` field MUST hash to `sd_hash`, after serialization to the `~` compact form.

bc-pi:
Such a field was removed in -06,`kb_jwt` in PR https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/344, when it became more evident that that the folks wanting to use JSON serialization were intending to do key binding by other means - some kind of enveloping rather than a separate KB JWT.  That along with the introduction of the _sd_hash to the KB JWT (which is much trickier than you've suggested esp wrt to the General JWS JSON Serialization) motivated the removal of the field.

bifurcation:
That just further emphasizes that we need to address #374  and separate "SD-JWT without KB" and "SD-JWT with KB", since the JSON serialization folks want to use the former and not the latter.  If we have that separation, this section can just have a JSON format for "SD-JWT without KB" and there's no mismatch.

danielfett:
The discussion about KB JWT in JSON-serialized JWS came up in the EUDI Wallet context as well, so we plan to fix it: https://github.com/oauth-wg/oauth-selective-disclosure-jwt/issues/403

PR #414, opened 2024-03-08
Rework JSON Serialization section

danielfett:
This addresses Issue #403 and #392 

Uses [this](https://github.com/openwallet-foundation-labs/sd-jwt-python/pull/19) new branch in sd-jwt-python.

Rendered version: https://drafts.oauth.net/oauth-selective-disclosure-jwt/danielfett/new-json-serialization/draft-ietf-oauth-selective-disclosure-jwt.html


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -765,35 +765,64 @@ Otherwise, the processed SD-JWT payload can be passed to the application to be u
...
 
-Using a payload similar to that from [Example 1](#example-1), the following is a non-normative example of
-a JWS JSON serialized SD-JWT from an Issuer with all the respective Disclosures.
+If a Key Binding JWT is present, the digest in the `sd_hash` claim MUST be taken
+over a string built as described in (#integrity-protection-of-the-presentation).
+The "Issuer-signed JWT" part is built by concatenating the protected header, the
+payload, and the signature of the JWS JSON serialized SD-JWT using a `.`
+character as a separator, and using the Disclosures from the `disclosures`
+member of the unprotected header. In case of multiple signatures, only the first
+one is used.
```
bc-pi, line 787: My first read of this last sentence had me thinking it was only about using the first signature value for constructing the "Issuer-signed JWT" part for input to the `sd_hash`. 

But I think it's also trying to say that, in the case of multiple signatures in the general serialization, the `disclosures` can only show up (and be used for constructing the `sd_hash` input) in the `header` of first of the `signatures` array.  And that the `kb_jwt` has to be there too.

And now I see that is what is said in the General JSON Serialization subsection below. Sorry!

But a bit more clarity here might still be good? I don't know that this is that much better but maybe:

```suggestion
one is used for the Disclosures and Key Binding JWT. 
```

Side note: using only the first one feels kinda icky TBH and the general issue was one reason I'd tried to avoid using the unprotected header for the SD-JWT pieces. But it makes sense to do it this way given all the factors and constraints involved.

bc-pi:
This one needs a doc history entry under -09 :)

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: Thanks, I think the clarification is useful! If we can come up with a better way than just using the first disclosure, I would be open to that.

danielfett:
> This one needs a doc history entry under -09 :)

Added!

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -765,35 +765,64 @@ Otherwise, the processed SD-JWT payload can be passed to the application to be u
...
+ * `kb_jwt`: A Key Binding JWT as described in (#kb-jwt).
 
-Using a payload similar to that from [Example 1](#example-1), the following is a non-normative example of
-a JWS JSON serialized SD-JWT from an Issuer with all the respective Disclosures.
+If a Key Binding JWT is present, the digest in the `sd_hash` claim MUST be taken
+over a string built as described in (#integrity-protection-of-the-presentation).
+The "Issuer-signed JWT" part is built by concatenating the protected header, the
+payload, and the signature of the JWS JSON serialized SD-JWT using a `.`
+character as a separator, and using the Disclosures from the `disclosures`
+member of the unprotected header. In case of multiple signatures, only the first
```
Sakurann, line 786: disclosures and paylaod need to be base64 url encoded before being concatenated with `.`, no?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -765,35 +765,64 @@ Otherwise, the processed SD-JWT payload can be passed to the application to be u
...
+
+<{{examples/json_serialization_general/sd_jwt_presentation.json}}
+
+## Verification of the JWS JSON Serialized SD-JWT
+
+Verification of the JWS JSON serialized SD-JWT follows the same rules defined in
+(#verification), except for the following aspects:
+
+ * The SD-JWT does not need to be split into component parts and the Disclosures
+   can be found in the respective member of the JSON object.
```
Sakurann, line 822: Disclosures can be found in the unprotected header, no? I am probably just not reading this right..

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: ok, I read unprotected header instead of protected header, all good.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
   can be found in the respective member of the JSON object in the unprotected header.
```

danielfett, line 822:
```suggestion
   can be found in the `disclosures` member of the unprotected header.
```

Sakurann (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -765,35 +765,64 @@ Otherwise, the processed SD-JWT payload can be passed to the application to be u
...
+
+The following is a non-normative example of a presentation of a JWS JSON
+serialized SD-JWT including a Key Binding JWT using the General JSON
+Serialization:
+
+<{{examples/json_serialization_general/sd_jwt_presentation.json}}
+
+## Verification of the JWS JSON Serialized SD-JWT
+
+Verification of the JWS JSON serialized SD-JWT follows the same rules defined in
```
danielfett, line 818:
```suggestion
Verification of the JWS JSON serialized SD-JWT follows the rules defined in
```

bc-pi (approved):

PR #435, opened 2024-06-12
exclusively use the cnf claim for enabling Key Binding

### LINKED ISSUE #419: why isn't cnf the only way to bind to a holder public key?
Right after the IETF 119 OAuth session yesterday in a talk with @selfissued and @ve7jtb that started with discussion of PR #394, the question again came up (with some surprise on their part) as to why the `cnf` claim isn't the one and only way specified to put the holder's public key in the issuer-signed JWT. I honestly had a hard time giving a good rational for SD-JWT not being more prescriptive with it's requirements here.

It seems that using the `cnf` claim exclusively would be more straightforward and much better for interop. 

@bifurcation has previously made similar comments, "it seems like there's an unnecessary level of ambiguity around how the Issuer-signed JWT binds to a holder public key.  Can we not just say `cnf`?  Why do we need more?  This point actually worries me more than [others] -- without clarity on this, there is no hope of writing stacks that interop."

Using the `cnf` claim exclusively could also help with some clarity in the [security considerations](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-08.html#section-10.6) where there currently has to be wording like "[if there's] no recognized Key Binding data is present in the SD-JWT..." and " ... the Issuer might have added the key to the SD-JWT in a format/claim that is not recognized by the Verifier.".

Note this is related to but separate from the question of if the presence of a `cnf` claim in the issuer-signed JWT means that the verifier should have to require a KB-JWT. The `cnf` claim can be made _the way_ to do holder binding while still leaving that up to verifier policy.

bc-pi:
More definitive language around the exclusive use of the cnf claim for enabling Key Binding

might just fix #419

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -996,16 +996,10 @@ credential, the type and contents of other credentials expected to be
 presented at the same time, etc.
 
 It is important that a Verifier does not make its security policy
-decisions based on data that can be influenced by an attacker or that
-can be misinterpreted. For this reason, when deciding whether Key
-Binding is required or not, Verifiers MUST NOT take into account:
-
- * whether the Holder has provided an SD-JWT+KB or a bare SD-JWT, since otherwise an
-   attacker could strip the KB-JWT from an SD-JWT+KB and present the resulting
-   SD-JWT, or
- * whether Key Binding data is present in the SD-JWT or not, as the
-   Issuer might have added the key to the SD-JWT in a format/claim that
-   is not recognized by the Verifier.
+decisions based on data that can be influenced by an attacker. For this reason, when deciding whether Key
+Binding is required or not, Verifiers MUST NOT take into account
+whether the Holder has provided an SD-JWT+KB or a bare SD-JWT, since otherwise an
+attacker could strip the KB-JWT from an SD-JWT+KB and present the resulting SD-JWT.
 
 If a Verifier determines that Key Binding is required for a
 particular use case and the Holder presents either a bare SD-JWT or an SD-JWT+KB with
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1001,6 +1001,8 @@ Binding is required or not, Verifiers MUST NOT take into account
 whether the Holder has provided an SD-JWT+KB or a bare SD-JWT, since otherwise an
 attacker could strip the KB-JWT from an SD-JWT+KB and present the resulting SD-JWT.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -105,7 +105,7 @@ values not disclosed in the Disclosures.
 This document also defines a format for SD-JWTs with Key Binding (SD-JWT+KB).
 By optionally sending an SD-JWT+KB to a
 Verifier, the Holder can prove to the Verifier that they hold the private key
-associated to the SD-JWT (e.g., using the `cnf` claim [@?RFC7800]). The strength of the binding is conditional upon the trust
```
Sakurann, line 108: i didn't know @? is valid

Sakurann (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: informational @? vs normative @! afaik

danielfett: `@?` equals `@` AFAIK

danielfett (approved):

bc-pi:
Need to look at text in one or more of the sec considerations too... todo

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -991,16 +996,10 @@ credential, the type and contents of other credentials expected to be
...
- * whether the Holder has provided an SD-JWT+KB or a bare SD-JWT, since otherwise an
-   attacker could strip the KB-JWT from an SD-JWT+KB and present the resulting
-   SD-JWT, or
- * whether Key Binding data is present in the SD-JWT or not, as the
-   Issuer might have added the key to the SD-JWT in a format/claim that
-   is not recognized by the Verifier.
+decisions based on data that can be influenced by an attacker. For this reason, when deciding whether Key
+Binding is required or not, Verifiers MUST NOT take into account
+whether the Holder has provided an SD-JWT+KB or a bare SD-JWT, since otherwise an
+attacker could strip the KB-JWT from an SD-JWT+KB and present the resulting SD-JWT.
```
danielfett, line 1002:
```suggestion
decisions based on data that can be influenced by an attacker. For this reason, when deciding whether Key
Binding is required or not, Verifiers MUST NOT take into account
whether the Holder has provided an SD-JWT+KB or a bare SD-JWT, since otherwise an
attacker could strip the KB-JWT from an SD-JWT+KB and present the resulting SD-JWT.

Furthermore, Verifiers should be aware that Key Binding information may have been added to an SD-JWT in a format that they do not recognize and therefore may not be able to tell whether the SD-JWT supports Key Binding or not.
```

PR #441, opened 2024-06-27
Remove two extraneous occurrences of "and without a KB-JWT"

### LINKED ISSUE #440: Mentions of KB-JWT that could be removed
The distinguishing of SD-JWT from SD-JWT+KB is a good clarifying change in -09. Some extra mentions of KB-JWT still remain that could probably removed:

In [5. ](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-09.html#section-5)[SD-JWT and SD-JWT+KB Data Formats](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-09.html#name-sd-jwt-and-sd-jwtkb-data-fo):

>An SD-JWT without Disclosures and without a KB-JWT

and

>An SD-JWT with Disclosures and without a KB-JWT

The "and without a KB-JWT" could be removed as there are separate examples for SD-JWT+KB and SD-JWT is now defined not to include KB-JWT.

bc-pi:
to fix #440

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -311,12 +311,12 @@ To further illustrate the SD-JWT format, the following examples show a few diffe
 SD-JWT permutations, both with and without various constituent parts.
 
 
-An SD-JWT without Disclosures and without a KB-JWT:
+An SD-JWT without Disclosures:
 ```
 <Issuer-signed JWT>~
 ```
 
-An SD-JWT with Disclosures and without a KB-JWT:
+An SD-JWT with Disclosures:
 ```
 <Issuer-signed JWT>~<Disclosure 1>~<Disclosure N>~
 ```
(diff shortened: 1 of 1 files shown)


srosenda (approved):
👍 looks good

Sakurann (approved):
I am not sure these are actually extraneous, but ok

srosenda:
> I am not sure these are actually extraneous, but ok

I was not completely sure either and that's why I posted only the issue and no PR first. I still think that in chapter 5. the extra mentions are confusing as the text and the ABNF are separating the SD-JWT and SD-JWT+KB formats.

In contrast, similar explicit mention [here](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-09.html#section-8.3-4.2) in the verification section is clarifying to me.

bc-pi:
I wasn't sure when first looking at it and actually thing it's okay either way but have come to think it's more okay this way.

PR #446, opened 2024-07-30
Clarification on composition of SD-JWT+KB

danielfett:
... as proposed by Paul. It is easy to read the current wording in a wrong way.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -265,7 +265,7 @@ An SD-JWT is composed of
 
 An SD-JWT+KB is composed of
 
-* an SD-JWT, and
+* an SD-JWT (i.e., an Issuer-signed JWT and zero or more Disclosures), and
 * a Key Binding JWT.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -265,7 +265,7 @@ An SD-JWT is composed of
 
 An SD-JWT+KB is composed of
 
-* an SD-JWT, and
+* an SD-JWT (i.e., an Issuer-signed JWT and zero or more Disclosures), and
 * a Key Binding JWT.
```
paulbastian, line 269:
```suggestion
* an Issuer-signed JWT,
* zero or more Disclosures, and
* a Key Binding JWT.
```
This is the most obvious to me.

danielfett: I proposed the above to clarify that the SD-JWT+KB contains an SD-JWT. But I'm fine with either.

bc-pi: I think the wording that Daniel is more accurate and reflective of what the intent of the +kb terminology.

PR #448, opened 2024-07-31
Strengthen privacy considerations

wbl:


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1317,28 +1317,12 @@ The following types of unlinkability are considered here:
...
-Where such callbacks are necessary, they MUST be executed in a manner that
-preserves privacy and does not disclose details about the credential to the Issuer. It is
-important to note that the timing of such requests could potentially serve as a side-channel.
-
-Verifier/Verifier unlinkability and presentation unlinkability can be achieved using batch issuance: A batch
-of credentials based on the same claims is issued to the Holder instead of just
-a single credential. The Holder can then use a different credential for each
-Verifier or even for each session with a Verifier. New key binding keys and
+SD-JWT MUST NOT be used when Issuer/Verifier unlinkability is required. This
+includes age verification on the Web. When users reveal their ages they do not
```
jogu, line 1321: I think "This includes age verification on the Web" is overly generic / too strong - I don't think we can say with certainty that all age verification use cases on the web must have issuer/verifier unlinkability (for example if the verifier already has other identity records for this user the additional risk seems very slow), and at the end of the day this is a balanced choice, particularly if the fallback (if an SD-JWT can't be used) is for the user to upload a picture of their full driving license which would be even worse for privacy. I would imagine the concerns can sometimes be solved by making it clear to the user what is happening.

bc-pi:
an alternative starting point is offered in #451

PR #452, opened 2024-08-12
tightening the exposition

Sakurann:
reflecting MikeJ's feedback

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -105,7 +105,7 @@ values not disclosed in the Disclosures.
 This document also defines a format for SD-JWTs with Key Binding (SD-JWT+KB).
 By optionally sending an SD-JWT+KB to a
 Verifier, the Holder can prove to the Verifier that they hold the private key
-associated with the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
+associated to the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
 in the protection of the private key of the key pair an SD-JWT is bound to.
 
 SD-JWT can be used with any JSON-based representation of claims, including JSON-LD.
@@ -118,7 +118,7 @@ wherever possible.
 
 This specification defines two primary data formats:
 
-1. SD-JWT is a composite structure enabling selective disclosure of its contents. It comprises of the following:
+1. SD-JWT is a composite structure enabling selective disclosure of its contents. It comprises the following:
   - A format for enabling selective disclosure in nested JSON data structures,
     supporting selectively disclosable object properties (name/value pairs) and array elements
   - A format for encoding the selectively disclosable data items
@@ -127,7 +127,7 @@ This specification defines two primary data formats:
   - An alternate format extending the JWS JSON Serialization, also allowing for
     transport of the Issuer-signed JSON data structure and disclosure data
 
-2. SD-JWT+KB is a composite structure enabling cryptographic key binding when presented to the Verifier. It comprises of the following:
+2. SD-JWT+KB is a composite structure enabling cryptographic key binding when presented to the Verifier. It comprises the following:
   - A facility for associating an SD-JWT with a key pair
   - A format for a Key Binding JWT (KB-JWT) that proves possession of the private key of
     the associated key pair
@@ -737,7 +737,7 @@ The following Disclosures are created by the Issuer:
 
 {{examples/simple/disclosures.md}}
 
-The payload is then signed by the Issuer to create a following JWT:
+The payload is then signed by the Issuer to create the following JWT:
 
 <{{examples/simple/sd_jwt_jws_part.txt}}
 
@@ -1007,7 +1007,7 @@ has not been tampered with since the issuance. The Issuer-signed JWT MUST be rej
 
 The security of the Issuer-signed JWT depends on the security of the signature algorithm.
 Any of the JWS asymmetric digital signature algorithms registered in [@IANA.JWS.Algorithms]
-that meet the security requirements described in the last paragraph of Section 5.2 of [RFC7515]
+that meet the security requirements described in the last paragraph of Section 5.2 of [@RFC7515]
 can be used, including post-quantum algorithms, when they are ready.
 
 ## Manipulation of Disclosures {#sec-disclosures}
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -108,7 +108,7 @@ Verifier, the Holder can prove to the Verifier that they hold the private key
 associated to the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
 in the protection of the private key of the key pair an SD-JWT is bound to.
 
-SD-JWT can be used with any JSON-based representation of claims, including JSON-LD.
+SD-JWT can be used with any JSON-based representation of claims.
 
 This specification aims to be easy to implement and to leverage
 established and widely used data formats and cryptographic algorithms
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -45,7 +45,7 @@ organization="Ping Identity"
 
 This specification defines a mechanism for selective disclosure of individual elements of a JSON object
 used as the payload of a JSON Web Signature (JWS) structure.
-It encompasses various applications, including but not limited to the selective disclosure of JSON Web Token (JWT) claims.
+It can be used for multiple applications, including but not limited to the selective disclosure of JSON Web Token (JWT) claims.
 
 {mainmatter}
 
@@ -53,7 +53,7 @@ It encompasses various applications, including but not limited to the selective 
 
 This document specifies conventions for creating JSON Web Signature (JWS) [@!RFC7515]
 structures with JSON [@!RFC8259] objects as the payload while supporting selective disclosure of individual elements of that JSON.
-Because JSON Web Token (JWT) [@!RFC7519] is a very prevalent application of JWS with a JSON payload, the selective disclosure of JWT claims receives primary treatment herein. However, that does not preclude the mechanism's applicability to other or more general applications of JWS with JSON payloads.
+Because JSON Web Token (JWT) [@!RFC7519] is a very prevalent application of JWS with a JSON payload, the selective disclosure of JWT claims receives primary treatment herein. However, that does not preclude the mechanism's applicability to other applications of JWS with JSON payloads.
 
 The JSON-based representation of claims in a signed JWT is
 secured against modification using JWS digital
@@ -76,23 +76,23 @@ the Verifier becomes crucial to ensure minimum disclosure and prevent
 Verifiers from obtaining claims irrelevant for the transaction at hand.
 SD-JWTs defined in this document enable such selective disclosure of JWT claims.
 
-One example of a multi-use JWT is a verifiable credential, an Issuer-signed
+One example of a multi-use JWT is an Issuer-signed
 credential that contains the claims about a subject, and whose authenticity can be
 cryptographically verified.
 
 Similar to the JWT specification on which it builds, this document is a product of the
-Web Authorization Protocol (oauth) working group. However, while both JWT and SD-JWT
+Web Authorization Protocol (OAuth) working group. However, while both JWT and SD-JWT
 have potential OAuth 2.0 applications, their utility and application is certainly not constrained to OAuth 2.0.
 JWT was developed as a general-purpose token format and has seen widespread usage in a
 variety of applications. SD-JWT is a selective disclosure mechanism for JWT and is
 similarly intended to be general-purpose specification.
 
 While JWTs with claims describing natural persons are a common use case, the
-mechanisms defined in this document can be used for other use cases as well.
+mechanisms defined in this document are also applicable to other use cases.
 
 In an SD-JWT, claims can be hidden, but cryptographically
 protected against undetected modification. "Claims" here refers to both
-object properties (name-value pairs) as well as array elements. When issuing the SD-JWT to
+object properties (name/value pairs) as well as array elements. When issuing the SD-JWT to
 the Holder, the Issuer includes the cleartext counterparts of all hidden
 claims, the so-called Disclosures, outside the signed part of the SD-JWT.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -105,7 +105,7 @@ values not disclosed in the Disclosures.
 This document also defines a format for SD-JWTs with Key Binding (SD-JWT+KB).
 By optionally sending an SD-JWT+KB to a
 Verifier, the Holder can prove to the Verifier that they hold the private key
-associated to the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
+associated with the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
```
bc-pi, line 108: I would ask that Mike's feedback not overwrite the phrasing that seemed more natural to Martin Thomson https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/394/files/d150ec24a8a369af4976c2739d8b97ac5650e603#r1416409939 

```suggestion
associated to the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -118,17 +118,17 @@ wherever possible.
 
 This specification defines two primary data formats:
 
-1. SD-JWT is a composite structure enabling selective disclosure of its contents. The definition in this document comprises the following:
+1. SD-JWT is a composite structure enabling selective disclosure of its contents. It comprises of the following:
```
bc-pi, line 121:
```suggestion
1. SD-JWT is a composite structure enabling selective disclosure of its contents. It comprises the following:
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -118,17 +118,17 @@ wherever possible.
...
+    supporting selectively disclosable object properties (name/value pairs) and array elements
   - A format for encoding the selectively disclosable data items
   - A format extending the JWS Compact Serialization, allowing for the combined
     transport of the Issuer-signed JSON data structure and the disclosable data items
   - An alternate format extending the JWS JSON Serialization, also allowing for
     transport of the Issuer-signed JSON data structure and disclosure data
 
-2. SD-JWT+KB is a composite structure enabling cryptographic key binding when presented to the Verifier. The definition in this document comprises the following:
-  - A facility for associating an SD-JWT to a key pair
+2. SD-JWT+KB is a composite structure enabling cryptographic key binding when presented to the Verifier. It comprises of the following:
```
bc-pi, line 130:
```suggestion
2. SD-JWT+KB is a composite structure enabling cryptographic key binding when presented to the Verifier. It comprises the following:
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -737,11 +737,11 @@ The following Disclosures are created by the Issuer:
 
 {{examples/simple/disclosures.md}}
 
-The payload is then signed by the Issuer to create a JWT like the following:
+The payload is then signed by the Issuer to create a following JWT:
```
bc-pi, line 740:
```suggestion
The payload is then signed by the Issuer to create the following JWT:
```

bc-pi (changes requested):
Thank you for putting this PR together! I made a few minor suggestions but otherwise it looks tight.

bc-pi:
> reflecting MikeJ's feedback

to be clear, the feedback was from Michael B. Jones

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett: Jeez, such an impactful change hidden away in this unconspicious PR... thanks, Brian, for catching that.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1007,6 +1007,7 @@ has not been tampered with since the issuance. The Issuer-signed JWT MUST be rej
 
 The security of the Issuer-signed JWT depends on the security of the signature algorithm.
 Any of the JWS asymmetric digital signature algorithms registered in [@IANA.JWS.Algorithms]
+that meet the security requirements described in the last paragraph of Section 5.2 of [RFC7515]
```
danielfett, line 1010:
```suggestion
that meet the security requirements described in the last paragraph of Section 5.2 of [@RFC7515]
```

danielfett (approved):
Approved with one suggestions.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: I am very sorry I did not catch that. you are absolutely right, reverting this.

bc-pi (approved):
looks good now, thanks for further tightening the exposition

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -935,8 +935,8 @@ unprotected header parameters are defined:
 In an SD-JWT+KB, `kb_jwt` MUST be present when using the JWS JSON Serialization,
 and the digest in the `sd_hash` claim MUST be taken over the SD-JWT as described
 in (#integrity-protection-of-the-presentation). This means that even when using
-the JWS JSON Serialization, the representation as a regular SD-JWT MUST be
-created temporarily to calculate the digest. In detail, the SD-JWT part is built
+the JWS JSON Serialization, the representation as a regular SD-JWT Compact Serialization MUST be
+created temporarily to calculate the digest. In detail, the SD-JWT Compact Serialization part is built
```
selfissued, line 939:
```suggestion
the JWS JSON Serialization, the representation as a regular SD-JWT using the Compact Serialization MUST be
created temporarily to calculate the digest. In detail, the SD-JWT using the Compact Serialization part is built
```

selfissued (approved):
All these changes look good.  Thanks for doing them.  By my count, this addresses 36 of my 59 review comments at https://mailarchive.ietf.org/arch/msg/oauth/5V6nVIfYw3BxsayLg1iY-YRUxdc/ - leaving 23.  Those that I believe have not yet been addressed are listed below.  I assume they will be addressed in subsequent PRs and/or discussed on list. 

I'm approving because I'm fine with this existing set being merged in now.  (Although a few of the remaining ones are also purely editorial, so could also be added. here.)

Section 1. [Introduction](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-introduction):  The usage of “Holder” in “used a number of times by the user (the "Holder" of the JWT)” inconsistent with its usage in the definitions, which defines “Holder” as being “an entity”.  The usage here in the introduction makes the holder into a person rather than an entity.  Please adjust the language here to not confuse the user who utilizes the holder with the holder itself.
Section 1. [Introduction](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-introduction):  In “"Claims" here refers to both object properties (name-value pairs) as well as array elements.” change “array elements” to “elements in arrays that are the values of name/value pairs” (or something like that).  Without saying what the array elements are, readers will be confused about what’s being referred to.  I’d apply this clarification in 4.1. [SD-JWT and Disclosures](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-sd-jwt-and-disclosures) and other applicable locations as well.
Section 1. [Introduction](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-introduction):  In “they hold the private key associated to the SD-JWT”, change “associated to” to “associated with”.
Section 1. [Introduction](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-introduction):  During IESG and/or RFC Editor review, we’re likely going to be asked to add a reference for JSON-LD at “SD-JWT can be used with any JSON-based representation of claims, including JSON-LD.”  If we don’t want that, I’d delete “, including JSON-LD” now.
Section 1.1. [Feature Summary](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-feature-summary):  In “A format for enabling selective disclosure in nested JSON data structures, supporting selectively disclosable object properties (name-value pairs) and array elements”, consider expanding “array elements” in the same manner as the preceding comment to make the meaning more evident.
Section 1.2. [Conventions and Terminology](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-conventions-and-terminology):  I suggest moving the “base64url” definition to the Terms and Definitions section and use a parallel structure to that at https://www.rfc-editor.org/rfc/rfc7519.html#section-2.  Specifically, say “The “base64url” term denotes the URL-safe base64 encoding without padding defined in Section 2 of [RFC7515].”  Then introduce the rest of the definitions with the phrase “These terms are defined by this specification:” as is done in RFC 7519.  The current presentation is fairly jarring.
Section 5.2.2. [Disclosures for Array Elements](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-disclosures-for-array-eleme):  Some readers and implementers will be confused at this point because the array disclosure doesn’t indicate what Claim Name the disclosed array element corresponds to.  Please say here how the Claim Name information is communicated.  At the very least, add a forward reference saying that the way the array element disclosure is associated with the Claim Name is described in 5.2.4.2. [Array Elements](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-array-elements).
Section 5.2.4.2. [Array Elements](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-array-elements):  I suggest adding an example also enabling disclosing a second array element, so that the syntax will be clear to readers.  In other words, show what “unless a matching Disclosure for the second element is received” would look like.  Or add a forward reference showing a place that it’s done.
Section 5.2.6. [Recursive Disclosures](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-recursive-disclosures):  Change “conceal presence” to “conceal the presence”.
Section 5.3. [Key Binding JWT](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-key-binding-jwt):  Change “MUST NOT be none.” to “It MUST NOT be "none".”.
Section 5.3.2. [Validating the Key Binding JWT](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-validating-the-key-binding-):  In “if the SD-JWT contains a cnf value with a jwk member, the Verifier could parse the provided JWK and use it to verify the Key Binding JWT.”, change “could” to “MUST”.  Make this normative to increase interoperability!
Section 6.1. [Issuance](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-issuance):  There are many places from here on where the label “SHA-256 Hash” is used, for instance “SHA-256 Hash: jsu9yVulwQQlhFlM_3JlzMaSFzglhQG0DpfayQwLUK4”.  Change all of these to “Base64url-Encoded SHA-256 Hash” for correctness.
Section 6.1. [Issuance](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-issuance):  Change “The following Key Binding JWT payload was created and signed for this presentation by the Holder:” to “The following Key Binding JWT Claims Set was created and signed for this presentation by the Holder:”, again, to use the standard JWT claims terminology.
Section 8.1. [Verification of the SD-JWT](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-verification-of-the-sd-jwt):  Delete “nbf” from “claims such as nbf, iat, and exp” and everywhere else in the spec, other than in 10.7. [Selectively-Disclosable Validity Claims](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-selectively-disclosable-val), where both “iat” and “nbf” should be listed, as described in my comment there.  “iat” (Issued At) is a standard validity claim in JWT tokens (for instance, ID Tokens), whereas “nbf” (Not Before) is rarely used, since it is only useful when future-dating tokens, which is rare.
Section 10.1. [Mandatory Signing of the Issuer-signed JWT](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-mandatory-signing-of-the-is):  …  Also, the post-quantum language seems superfluous and non-actionable, so I’d delete it.
Section 10.7. [Selectively-Disclosable Validity Claims](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-selectively-disclosable-val):  Add “iat (Issued At) to the validity claims list before “nbf”, and change “nbf (Not Before)” to “nbf (Not Before) (if present)”.  “iat” (Issued At) is a standard validity claim in JWT tokens (for instance, ID Tokens), whereas “nbf” (Not Before) is rarely used, since it is only useful when future-dating tokens, which is rare.  Change all uses of “nbf” to “iat” elsewhere in the spec.
Section 11.1. [Storage of Signed User Data](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-storage-of-signed-user-data):  Change “OpenID Connect” to “OpenID Connect [OpenID.Core]”.
Section 12. [Acknowledgements](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-acknowledgements):  Consider sorting the acknowledgements alphabetically by last name, which is the usual convention.
Section 14.2. [Informative References](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-informative-references):  Update the reference “[I-D.ietf-oauth-sd-jwt-vc]” to https://datatracker.ietf.org/doc/html/draft-ietf-oauth-sd-jwt-vc-04.
Section A.1. [Simple Structured SD-JWT](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-simple-structured-sd-jwt): Change “allowing to separately disclose individual members of the claim” to “allowing separate disclosure of individual members of the claim”.
Section A.5. [Elliptic Curve Key Used in the Examples](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-elliptic-curve-key-used-in-): Change “The public key used to validate a Key Binding JWT can be found in the examples as the content of the cnf claim.” to “The public key used to validate a Key Binding JWT can be found in the examples as the value of the jwk member of the cnf claim.”
Section Appendix B. [Disclosure Format Considerations](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-disclosure-format-considera): At the end of the paragraph introduced by “1. Canonicalization”, consider adding the sentence “Canonicalization schemes have a long history of creating interoperability problems in practice.”
Everywhere:  Consider changing the name “…” to something more indicative of its function, such as “_sd_element” or “_sd_item”.  Or alternatively, at least provide rationale for why that non-obvious name was chosen.

bc-pi:
> Approved with one suggestions.

how many suggestion**s**?

Sakurann:
> Those that I believe have not yet been addressed are listed below. I assume they will be addressed in subsequent PRs and/or discussed on list.

the remaining ones are not addressed for certain reason - there is misunderstanding, editors believe the current way is better for good reasons, etc. I will be responding to the ML with the explanations.

PR #454, opened 2024-08-22
additional updates from Mike's review

bc-pi:
see https://mailarchive.ietf.org/arch/msg/oauth/cPD_LWWEPedQR3SiU17Juamtwt8/ and https://mailarchive.ietf.org/arch/msg/oauth/ulgKG6F-_RU6S2TaocZpbqEFgzs/

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -669,7 +669,7 @@ The Key Binding JWT MUST be a JWT according to [@!RFC7519] and its payload MUST 
 
 * in the JOSE header,
     * `typ`: REQUIRED. MUST be `kb+jwt`, which explicitly types the Key Binding JWT as recommended in Section 3.11 of [@!RFC8725].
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -174,7 +174,7 @@ Issuer:
 :  An entity that creates SD-JWTs.
 
 Holder:
-:  An entity that received SD-JWTs from the Issuer and has control over them.
+:  An entity that received SD-JWTs from the Issuer and has control over them. The term may refer to the actual user, the supporting hardware and software in their possession, or both.
```
danielfett, line 177:
```suggestion
:  An entity that received SD-JWTs from the Issuer and has control over them. In the context of this document, the term may refer to the actual user, the supporting hardware and software in their possession, or both.
```

selfissued (approved):
Thanks for these additional updates.

bc-pi:
oh wow... I apparently didn't confirm the merge so this is still pending. I really thought it was done and published -11. Sigh. Sorry. Will look at doing a -12 soonish.

bc-pi:
see #455

PR #472, opened 2024-10-04
WGLC updates 1/x

bc-pi:
updates per https://mailarchive.ietf.org/arch/msg/oauth/QpovghcXuKkhO6MW93pOTyAsmlU/
and https://mailarchive.ietf.org/arch/msg/oauth/rzqD_lSqnY7MsWtP_1TCIyH8zyg/


danielfett (approved):
Thanks!

Sakurann (approved):

PR #473, opened 2024-10-04
 WGLC updates 2/x (remove sentence in intro)

bc-pi:
remove repetitive repetitive sentence per Dick https://mailarchive.ie)tf.org/arch/msg/oauth/brjnGB53MApGNCsXC03ungeWxI4/

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -108,8 +108,6 @@ Verifier, the Holder can prove to the Verifier that they hold the private key
 associated to the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
 in the protection of the private key of the key pair an SD-JWT is bound to.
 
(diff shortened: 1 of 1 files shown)


danielfett (approved):
Merci!

Sakurann (changes requested):
I am ok with adding Dick as a contributor, but I am not ok removing this sentence (line 111), sorry. like, this is the only sentence that is left from when we used to say "sd-jwt can be used with json-ld", and "The JSON-based representation of claims in a signed JWT..." in the intro is not as clear as the sentence being deleted.

bc-pi:
> I am ok with adding Dick as a contributor, but I am not ok removing this sentence (line 111), sorry. like, this is the only sentence that is left from when we used to say "sd-jwt can be used with json-ld", and "The JSON-based representation of claims in a signed JWT..." in the intro is not as clear as the sentence being deleted.

Okay, let's just close this one then.

bc-pi:
@danielfett is going to rework the introduction too

PR #475, opened 2024-10-05
 ISO/IEC 29100 is too private 

bc-pi:
https://standards.iso.org/ittf/PubliclyAvailableStandards/index.html no longer has anything for 29100 and https://www.iso.org/standard/45123.html indicates that ISO/IEC 29100:2011 is "Withdrawn" and "Revised by" ISO/IEC 29100:2024 that is now available for the low low special price of CHF 129, which is approximately $150 / €137 too much for me to be comfortable referencing

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1211,8 +1211,6 @@ Appropriate key management is essential, as any compromise can lead to unauthori
 
 ## Unlinkability {#unlinkability}
 
-The privacy principles of [@ISO.29100] should be adhered to.
-
 Unlinkability is a property whereby adversaries are prevented from correlating
 credential presentations of the same user beyond the user's consent.
 Without unlinkability, an adversary might be able to learn more about the user than the user
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1219,8 +1219,6 @@ The definition of `typ` in Section 4.1.9 of [@!RFC7515] recommends that the "app
 
 # Privacy Considerations {#privacy_considerations}
 
-The privacy principles of [@ISO.29100] should be adhered to.
-
 ## Unlinkability
 
 Unlinkability is a property whereby adversaries are prevented from correlating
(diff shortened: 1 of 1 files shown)


danielfett (approved):
Takk!

bc-pi:
> Takk!

That was the first one I didn't recognize. Google says Norwegian, è corretto?

danielfett:
Det er riktig.

Sakurann:
Based on Nat's comment, I don't think we need to remove this.

sakimura:
Just got notified that it is now available from https://standards.iso.org/ittf/PubliclyAvailableStandards/ .

Sakurann (changes requested):
per Nat's comment the standard is publicly available again :)

bc-pi:
Screenshot 2024-10-15 at 6 10 46 AM

Screenshot 2024-10-15 at 6 11 02 AM

bc-pi:
Screenshot 2024-10-17 at 5 55 59 AM

I don't think the ISO.29100 reference should have been included in the first place and it's unavailability isn't helping my opinion.

bc-pi:
https://mailarchive.ietf.org/arch/msg/oauth/SmGW8DevLyuydXAWLPeIgxX4MBo/ mentions this reference again

Sakurann (changes requested):
I believe what needs to be done is change the reference to 29100:2024 from 29100:2011, not removing the reference

bc-pi:
> I believe what needs to be done is change the reference to 29100:2024 from 29100:2011, not removing the reference

That's an option too. But have you read all of ISO/IEC 29100:2024 and are all of it's principles applicable to SD-JWT? I don't know, I've only glanced at it. But I don't personally find content like "The privacy principles of [@ISO.29100] should be adhered to." to be particularity helpful in a document like this and think it'd be better off without a vague and general statement to go see this other document that's not easily accessed. That how it's not easily accessed is more a reminder that it, in my opinion anyway, never should have been referenced from SD-JWT in the first place.

danielfett:
29100 is now available again, it seems.

I share the concern that the current reference is very broad and not actionable at all.

Sakurann (approved):
I give in to the concern that `current reference is very broad and not actionable at all`.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1209,10 +1209,10 @@ Appropriate key management is essential, as any compromise can lead to unauthori
 
 # Privacy Considerations {#privacy_considerations}
 
-The privacy principles of [@ISO.29100] should be adhered to.
-
 ## Unlinkability {#unlinkability}
 
+The privacy principles of [@ISO.29100] should be adhered to.
+
```
Sakurann, line 1215:
```suggestion
```

PR #467, opened 2024-09-21
introduction rewrite

dickhardt:
provide more context on what the problem is and why this work is happening, with a summary of how it works

deleted spurious language that reads like a patent application

PR #477, opened 2024-10-10
Rewrite introduction and abstract

danielfett:
Takes inspiration from #467 but is essentially a rewrite of the introduction from scratch.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -43,75 +43,37 @@ organization="Ping Identity"
 
 .# Abstract
 
-This specification defines a mechanism for selective disclosure of individual elements of a JSON object
-used as the payload of a JSON Web Signature (JWS) structure.
-It can be used for multiple applications, including but not limited to the selective disclosure of JSON Web Token (JWT) claims.
+This specification defines a mechanism for the selective disclosure of individual elements of a JSON-encoded data structure used as the payload of a JSON Web Signature (JWS).
+The primary use case is the selective disclosure of JSON Web Token (JWT) claims.
 
 {mainmatter}
 
 # Introduction {#Introduction}
 
-This document specifies conventions for creating JSON Web Signature (JWS) [@!RFC7515]
-structures with JSON [@!RFC8259] objects as the payload while supporting selective disclosure of individual elements of that JSON.
-Because JSON Web Token (JWT) [@!RFC7519] is a very prevalent application of JWS with a JSON payload, the selective disclosure of JWT claims receives primary treatment herein. However, that does not preclude the mechanism's applicability to other applications of JWS with JSON payloads.
+JSON-encoded data structures for exchange between systems are often secured against modification using JSON Web Signatures (JWS) [@!RFC7515].
+A popular application of JWS are JSON Web Tokens (JWT) [@!RFC7519], which are often used to represent a user's identity.
+An ID Token as defined in OpenID Connect [@?OpenID.Core], for example, is a JWT containing the user's claims created by the server for consumption by a client.
+In case the JWS is sent directly from the server to the client, as in OpenID Connect, the server can select which user claims to include in the JWT, minimizing the information shared with the client.
 
-The JSON-based representation of claims in a signed JWT is
-secured against modification using JWS digital
-signatures. A consumer of a signed JWT that has validated the
-signature can safely assume that the contents of the token have not been
-modified.  However, anyone receiving an unencrypted JWT can read all the
-claims. Likewise, anyone with the decryption key receiving encrypted JWT
-can also read all the claims.
+A new model is emerging that decouples the issuance of a JWT from its presentation.
+In this model, a JWT containing many claims is issued to an intermediate party, who holds the JWT (the Holder).
+The Holder, typically an end-user, can then present the JWT to different verifying parties (Verifiers), who each may only require a subset of the claims in the JWT.
+For example, the JWT may contain claims representing both an address and a birthdate.
+The Holder would like to disclose only their address to one Verifier, and only their birthdate to a different Verifier.
 
-One of the common use cases of a signed JWT is representing a user's
-identity. As long as the signed JWT is one-time
-use, it typically only contains those claims the user has consented to
-disclose to a specific Verifier. However, there is an increasing number
-of use cases where a signed JWT is created once and then used a number
-of times by the user (the "Holder" of the JWT). In such use cases, the signed JWT needs
-to contain the superset of all claims the user of the
-signed JWT might want to disclose to Verifiers at some point. The
-ability to selectively disclose a subset of these claims depending on
-the Verifier becomes crucial to ensure minimum disclosure and prevent
-Verifiers from obtaining claims irrelevant for the transaction at hand.
-SD-JWTs defined in this document enable such selective disclosure of JWT claims.
+This model requires a mechanism enabling selective disclosure of data elements while ensuring that Verifiers can still check the authenticity of the data provided.
+This specification defines such a mechanism for JSON-encoded payloads of JSON Web Signatures, with a primary use case being JSON Web Tokens.
 
-One example of a multi-use JWT is an Issuer-signed
-credential that contains the claims about a subject, and whose authenticity can be
-cryptographically verified.
+SD-JWT is based on an aproach called "salted hashes": For any data element that should be selectively disclosable, the Issuer of the SD-JWT does not include the cleartext of the data in the JSON-encoded payload of the JWS structure; instead, a hash digest of the data takes its place.
+For presentation to a Verifier, the Holder sends the signed payload along with with the cleartext of those claims it wants to disclose.
+The Verifier can then compute the digest of the cleartext data and confirm it is included in the signed payload.
+To ensure that Verifiers cannot guess cleartext values of non-disclosed data elements, an additional salt value is used when creating the digest and sent along with the cleartext when disclosing it.
 
-Similar to the JWT specification on which it builds, this document is a product of the
-Web Authorization Protocol (OAuth) working group. However, while both JWT and SD-JWT
-have potential OAuth 2.0 applications, their utility and application is certainly not constrained to OAuth 2.0.
-JWT was developed as a general-purpose token format and has seen widespread usage in a
-variety of applications. SD-JWT is a selective disclosure mechanism for JWT and is
-similarly intended to be a general-purpose specification.
+To prevent attacks in which an SD-JWT is presented to a Verifier without the Holder's consent, this specification additionally defines a mechanism for binding the SD-JWT to a key under the control of the Holder (Key Binding).
+When Key Binding is enforced, a Holder has to proof possession of a private key belonging to a public key contained in the SD-JWT itself.
+It usually does so by signing over a data structure containing transaction-specific data, here defined as the Key Binding JWT.
+An SD-JWT with a Key Binding JWT is called SD-JWT+KB in this specification.
 
-While JWTs with claims describing natural persons are a common use case, the
-mechanisms defined in this document are also applicable to other use cases.
-
-In an SD-JWT, claims can be hidden, but cryptographically
-protected against undetected modification. When issuing the SD-JWT to
-the Holder, the Issuer includes the cleartext counterparts of all hidden
-claims, the so-called Disclosures, outside the signed part of the SD-JWT.
-
-The Holder decides which claims to disclose to a particular Verifier and includes the respective
-Disclosures in the SD-JWT to that Verifier. The Verifier
-has to verify that all disclosed claim values were part of the
-Issuer-signed JWT. The Verifier will not, however, learn any hidden claim
-values not disclosed in the Disclosures.
-
-This document also defines a format for SD-JWTs with Key Binding (SD-JWT+KB).
-By optionally sending an SD-JWT+KB to a
-Verifier, the Holder can prove to the Verifier that they hold the private key
-associated to the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
-in the protection of the private key of the key pair an SD-JWT is bound to.
-
-SD-JWT can be used with any JSON-based representation of claims.
-
-This specification aims to be easy to implement and to leverage
-established and widely used data formats and cryptographic algorithms
-wherever possible.
 
 ## Feature Summary
 


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
-ability to selectively disclose a subset of these claims depending on
-the Verifier becomes crucial to ensure minimum disclosure and prevent
-Verifiers from obtaining claims irrelevant for the transaction at hand.
-SD-JWTs defined in this document enable such selective disclosure of JWT claims.
-
-One example of a multi-use JWT is an Issuer-signed
-credential that contains the claims about a subject, and whose authenticity can be
-cryptographically verified.
-
-Similar to the JWT specification on which it builds, this document is a product of the
```
danielfett, line 83: I dropped this part. If we want to keep it, I suggest including it under a separate "## Scope" headline below the intro or below the feature list.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
+
+SD-JWT is based on an aproach called "salted hashes": For any data element that should be selectively disclosable, the Issuer of the SD-JWT does not include the cleartext of the data in the JSON-encoded payload of the JWS structure; instead, a hash digest of the data takes its place.
+For presentation to a Verifier, the Holder sends the signed payload along with with the cleartext of those claims it wants to disclose.
+The Verifier can then compute the digest of the cleartext data and confirm it is included in the signed payload.
+To ensure that Verifiers cannot guess cleartext values of non-disclosed data elements, an additional salt value is used when creating the digest and sent along with the cleartext when disclosing it.
+
+To prevent attacks in which an SD-JWT is presented to a Verifier without the Holder's consent, this specification additionally defines a mechanism for binding the SD-JWT to a key under the control of the Holder (Key Binding).
+When Key Binding is enforced, a Holder has to proof possession of a private key belonging to a public key contained in the SD-JWT itself.
```
ju-cu, line 73:
```suggestion
When Key Binding is enforced, a Holder has to prove possession of a private key belonging to a public key contained in the SD-JWT itself.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
-
-SD-JWT can be used with any JSON-based representation of claims.
-
-This specification aims to be easy to implement and to leverage
-established and widely used data formats and cryptographic algorithms
-wherever possible.
+JSON-encoded data structures for exchange between systems are often secured against modification using JSON Web Signatures (JWS) [@!RFC7515].
+A popular application of JWS are JSON Web Tokens (JWT) [@!RFC7519], which are often used to represent a user's identity.
+An ID Token as defined in OpenID Connect [@?OpenID.Core], for example, is a JWT containing the user's claims created by the server for consumption by a client.
+In case the JWS is sent directly from the server to the client, as in OpenID Connect, the server can select which user claims to include in the JWT, minimizing the information shared with the client.
```
ju-cu, line 56: Isn't the important fact here that the client is also the party verifying the JWT? I think, adding this detail may be useful to understand the issuer-holder-verifier model defined afterwards as it may otherwise be hard for the reader to understand the fundamental shift.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
+
+A new model is emerging that decouples the issuance of a JWT from its presentation.
```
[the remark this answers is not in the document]
Sakurann: the AS that issues the access token and the RS that consumes it are assumed to be the same entity though, right?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
+In this model, a JWT containing many claims is issued to an intermediate party, who holds the JWT (the Holder).
+The Holder, typically an end-user, can then present the JWT to different verifying parties (Verifiers), who each may only require a subset of the claims in the JWT.
+For example, the JWT may contain claims representing both an address and a birthdate.
+The Holder would like to disclose only their address to one Verifier, and only their birthdate to a different Verifier.
+
+This model requires a mechanism enabling selective disclosure of data elements while ensuring that Verifiers can still check the authenticity of the data provided.
```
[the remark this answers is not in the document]
Sakurann: agree that it should be more explicit that selective disclosure becomes an important privacy feature because the same JWT can be used for different purposes.

draft-ietf-oauth-selective-disclosure-jwt.md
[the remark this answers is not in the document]
ju-cu: Right. I was thinking in technical terms (server and client communication) while the decoupling happens on a different layer which actually is a new (trust) model. I believe, it would be good to focus more on that change of trust where the RS/Verifier is not a single trusted entity (and certainly not the same entity as the one issuing the JWT) but any third party.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
-in the protection of the private key of the key pair an SD-JWT is bound to.
-
-SD-JWT can be used with any JSON-based representation of claims.
-
-This specification aims to be easy to implement and to leverage
-established and widely used data formats and cryptographic algorithms
-wherever possible.
+JSON-encoded data structures for exchange between systems are often secured against modification using JSON Web Signatures (JWS) [@!RFC7515].
+A popular application of JWS are JSON Web Tokens (JWT) [@!RFC7519], which are often used to represent a user's identity.
+An ID Token as defined in OpenID Connect [@?OpenID.Core], for example, is a JWT containing the user's claims created by the server for consumption by a client.
```
Sakurann, line 55:
```suggestion
A popular application of JWS are JSON Web Tokens (JWT) [@!RFC7519], which can be used to represent a user's identity.
For example, an ID Token as defined in OpenID Connect [@?OpenID.Core], is a JWT containing the user's identity claims created by a server for consumption by a client.
```
not sure user's identity is the most used application of JWTs..
is it JWT that is an application of JWS, or JWS that is an application of JWT? or they are not applications for either..?
i am not good in articles but why `the` server but `a` client?

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
In case a JWT is sent directly from a server to a client, as in OpenID Connect, the server can select real-time which user claims to include in the JWT, minimizing the information shared with the client, who validates the JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
 
 .# Abstract
 
-This specification defines a mechanism for selective disclosure of individual elements of a JSON object
-used as the payload of a JSON Web Signature (JWS) structure.
-It can be used for multiple applications, including but not limited to the selective disclosure of JSON Web Token (JWT) claims.
+This specification defines a mechanism for the selective disclosure of individual elements of a JSON-encoded data structure used as the payload of a JSON Web Signature (JWS).
```
bc-pi, line 46: I'm curious to understand why "JSON-encoded data structure" is used here and elsewhere in the PR vs. "JSON object" or "JSON document" or "bit o' JSON" or something? It reads a bit off to my eye, at first anyway, but I wanted to ask before suggesting changes or otherwise fighting about it.

danielfett: Mainly to avoid "JSON object" as we can also SD the elements of an array.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: I wrote some of that in response to questions/comments by Hannes about the appropriateness of this work in the OAUTH WG. Maybe the relationship to OAuth and JWT stuff isn't needed anymore. And the generally applicable bit is pretty much implied, in my opinion, even if Dick didn't see it.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi, line 55: "client" isn't wrong but has overloaded meaning and isn't used anywhere in this document prior to this PR 
```suggestion
An ID Token as defined in OpenID Connect [@?OpenID.Core], for example, is a JWT containing the user's claims created by the server for consumption by a relying party.
```

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: reflecting (maybe) prior comments along with my own nitpicks to create this frankenstein of a sentence : 
```suggestion
In cases where the JWT is sent immediately from the server to the relying party, as in OpenID Connect, the server can select at the time of issuance which user claims to include in the JWT, minimizing the information shared with the relying party, who validates the JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: Ah ok. I'd read individual elements of a "JSON document" or "bit o' JSON" to mean all the stuff that might be inside that thing like objects or elements of an array. But I take your point or intention.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
+A popular application of JWS are JSON Web Tokens (JWT) [@!RFC7519], which are often used to represent a user's identity.
+An ID Token as defined in OpenID Connect [@?OpenID.Core], for example, is a JWT containing the user's claims created by the server for consumption by a client.
+In case the JWS is sent directly from the server to the client, as in OpenID Connect, the server can select which user claims to include in the JWT, minimizing the information shared with the client.
+
+A new model is emerging that decouples the issuance of a JWT from its presentation.
+In this model, a JWT containing many claims is issued to an intermediate party, who holds the JWT (the Holder).
+The Holder, typically an end-user, can then present the JWT to different verifying parties (Verifiers), who each may only require a subset of the claims in the JWT.
+For example, the JWT may contain claims representing both an address and a birthdate.
+The Holder would like to disclose only their address to one Verifier, and only their birthdate to a different Verifier.
```
bc-pi, line 62: maybe... i dunno... 
```suggestion
The Holder may elect to disclose only the address to one Verifier, and only the birthdate to a different Verifier.
```

draft-ietf-oauth-selective-disclosure-jwt.md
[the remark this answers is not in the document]
bc-pi: @ju-cu is not wrong here 
I've long made the complaint that the new "three-party model" sounds silly because we've had 3 party models for a long time.
But not sure how much of that can or should be captured or conveyed here.
And the previous intro might have kind of captured that but not particularly well IMO 
what about a bit o' qualification to this sentence?

```suggestion
A new model is emerging that more fully decouples the issuance of a JWT from its presentation.
```

draft-ietf-oauth-selective-disclosure-jwt.md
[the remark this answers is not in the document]
bc-pi: something like this?
```suggestion
Privacy principles of minimal disclosure in conjunction with this model demand a mechanism enabling selective disclosure of data elements while ensuring that Verifiers can still check the authenticity of the data provided.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
+For presentation to a Verifier, the Holder sends the signed payload along with with the cleartext of those claims it wants to disclose.
+The Verifier can then compute the digest of the cleartext data and confirm it is included in the signed payload.
+To ensure that Verifiers cannot guess cleartext values of non-disclosed data elements, an additional salt value is used when creating the digest and sent along with the cleartext when disclosing it.
+
+To prevent attacks in which an SD-JWT is presented to a Verifier without the Holder's consent, this specification additionally defines a mechanism for binding the SD-JWT to a key under the control of the Holder (Key Binding).
+When Key Binding is enforced, a Holder has to proof possession of a private key belonging to a public key contained in the SD-JWT itself.
+It usually does so by signing over a data structure containing transaction-specific data, here defined as the Key Binding JWT.
```
bc-pi, line 74:
```suggestion
It usually does so by signing over a data structure containing transaction-specific data, herein defined as the Key Binding JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -43,75 +43,37 @@ organization="Ping Identity"
...
-associated to the SD-JWT (i.e., using the `cnf` claim [@!RFC7800]). The strength of the binding is conditional upon the trust
-in the protection of the private key of the key pair an SD-JWT is bound to.
-
-SD-JWT can be used with any JSON-based representation of claims.
-
-This specification aims to be easy to implement and to leverage
-established and widely used data formats and cryptographic algorithms
-wherever possible.
+JSON-encoded data structures for exchange between systems are often secured against modification using JSON Web Signatures (JWS) [@!RFC7515].
+A popular application of JWS are JSON Web Tokens (JWT) [@!RFC7519], which are often used to represent a user's identity.
```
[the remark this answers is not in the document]
danielfett: a```suggestion
A popular application of JWS is JSON Web Token (JWT) [@!RFC7519], a format that is often used to represent a user's identity.
```

File: draft-ietf-oauth-selective-disclosure-jwt.md (hunk 3 above)
Comment (Sakurann) [reply draft-ietf-oauth-selective-disclosure-jwt.md:56]:
```suggestion
In cases where the JWT is sent immediately from the server to the relying party, as in OpenID Connect, the server can select at the time of issuance which user claims to include in the JWT, minimizing the information shared with the relying party who validates the JWT.
```

draft-ietf-oauth-selective-disclosure-jwt.md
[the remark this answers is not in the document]
Sakurann:
```suggestion
A new model is emerging that fully decouples the issuance of a JWT from its presentation.
```

Sakurann (approved):

bc-pi (approved):

PR #478, opened 2024-10-16
Added comment regarding holder guarantees (Addressing Issue #470)

### LINKED ISSUE #470: Decoy Digest Implications
We (Guido Schmitz and Alex Hodder-Williams) have been working on producing a formal analysis of the current specification. We will share more details about this analysis soon, but wanted to point out some observations.

During the issuance phase, it is unclear to us how the holder gets guarantees regarding the entire set of disclosures that verify under the received issuer-signed JWT.
As specified: the holder should accept an issuance if the signature verifies, and the received set of disclosures all verify according to this SD-JWT presentation. However, the holder doesn't have guarantees that the digests in the issuer-signed JWT that they didn't receive a corresponding disclosure for are in fact decoy digests. These digests could contain other disclosures, which the holder would be unaware of. This would allow colluding issuers and verifiers to create a non-interactive covert channel, where the verifiers are looking for a specific digest. This seems to relate to 11.1 in the specification, but may be worth additional discussion in the document.

One possible technical solution could look as follows:
- The holder provides the issuer with a set of nonces that can be used as decoy digests.
- The issuer must then hash these nonces (this prevents the holder from inserting their own non-issued claims)
- The holder can then check if the digests they don't receive disclosures for match their original nonces.
Clearly this adds overhead and slightly complicates the standard, however, it is our view that the holder of a credential that has been issued to (and potentially about) them, should have full knowledge of what could be contained in that credential.

Another issue stemming from decoy digests, with a different consequence, is that if disclosures get dropped off the end of an SD-JWT presentation in the issuance phase, yet the presentation still verifies, the holder won't ever have the full set of disclosures, but will be unaware that they are missing information.
This may be mitigated at a different level. Any kind of authenticated channel between the issuer and holder prevents tampering of the initial set of disclosures. This may be out of scope of the SD-JWT document as it is likely to be protocol specific; but again, this might be worth including a remark about.

AlexHodder:
Resolves #470 by adding a comment to Section 8.1 (Verification of the SD-JWT) regarding holder guarantees after validating an SD-JWT presentation from an issuer.

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -869,7 +869,7 @@ an SD-JWT:
 5. If any Disclosure was not referenced by digest value in the Issuer-signed JWT (directly or recursively via other Disclosures), the SD-JWT MUST be rejected.
 6. Check that the SD-JWT is valid using claims such as `nbf`, `iat`, and `exp` in the processed payload. If a required validity-controlling claim is missing (see (#sd-validity-claims)), the SD-JWT MUST be rejected.
 
-If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose. 
+If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose.
 
 Note that these processing steps do not yield any guarantees to the Holder about having received a complete set of Disclosures. That is, for some digest values in the Issuer-signed JWT (which are not decoy digests) there may be no corresponding Disclosures, for example, if the message from the Issuer was truncated.
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -869,7 +869,7 @@ an SD-JWT:
 5. If any Disclosure was not referenced by digest value in the Issuer-signed JWT (directly or recursively via other Disclosures), the SD-JWT MUST be rejected.
 6. Check that the SD-JWT is valid using claims such as `nbf`, `iat`, and `exp` in the processed payload. If a required validity-controlling claim is missing (see (#sd-validity-claims)), the SD-JWT MUST be rejected.
 
-If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose.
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -869,7 +869,7 @@ an SD-JWT:
...
 6. Check that the SD-JWT is valid using claims such as `nbf`, `iat`, and `exp` in the processed payload. If a required validity-controlling claim is missing (see (#sd-validity-claims)), the SD-JWT MUST be rejected.
 
-If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose.
+If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose. Note that these processing steps do not yield any guarantees to the Holder about having received a complete set of disclosures. That is, there might exist digest values in the Issuer-signed JWT which are not decoy digests but correspond to disclosures not contained in the list of disclosures provided in the SD-JWT.
```
danielfett, line 872:
```suggestion
If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose. Note that these processing steps do not yield any guarantees to the Holder about having received a complete set of disclosures. That is, for some digest values in the Issuer-signed JWT (which are not decoy digests) there may be no corresponding disclosures.
```

Sakurann (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -869,7 +869,9 @@ an SD-JWT:
...
 6. Check that the SD-JWT is valid using claims such as `nbf`, `iat`, and `exp` in the processed payload. If a required validity-controlling claim is missing (see (#sd-validity-claims)), the SD-JWT MUST be rejected.
 
-If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose.
+If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose. 
+
+Note that these processing steps do not yield any guarantees to the Holder about having received a complete set of Disclosures. That is, for some digest values in the Issuer-signed JWT (which are not decoy digests) there may be no corresponding Disclosures, for example, if the message from the Issuer was truncated.
```
Sakurann, line 874:
```suggestion
If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Otherwise, the JSON document resulting from the preceding processing and verification steps, herein referred to as the processed SD-JWT payload, can be made available to the application to be used for its intended purpose.

Note that these processing steps do not yield any guarantees to the Holder about having received a complete set of Disclosures. That is, for some digest values in the Issuer-signed JWT (which are not decoy digests) there may be no corresponding Disclosures, for example, if the message from the Issuer was truncated.
```

bc-pi (approved):

PR #481, opened 2024-10-25
Address Denis' comments

danielfett:


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -204,7 +204,7 @@ The Key Binding JWT encodes a signature by the Holder's private key over
 
 * a hash of the SD-JWT,
 * a nonce to ensure the freshness of the signature, and
-* an audience value to indicate the intended audience for the document.
+* an audience value to indicate the intended Verifier or Verifiers for the document.
```
bc-pi, line 207: while "Verifier or Verifiers" is true I guess, I don't think we need to mention the plural here. 
```suggestion
* an audience value to indicate the intended Verifier for the document.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -137,7 +137,7 @@ Issuer:
 :  An entity that creates SD-JWTs.
 
 Holder:
-:  An entity that received SD-JWTs from the Issuer and has control over them. In the context of this document, the term may refer to the actual user, the supporting hardware and software in their possession, or both.
+:  An entity that received SD-JWTs from the Issuer and has control over them. In the context of this document, the term usually refers to the actual user and the supporting hardware and software in their possession.
```
bc-pi, line 140: I am not totally sure this change is needed or an improvement... ?

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -833,16 +835,14 @@ If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Othe
 
 Note that these processing steps do not yield any guarantees to the Holder about having received a complete set of Disclosures. That is, for some digest values in the Issuer-signed JWT (which are not decoy digests) there may be no corresponding Disclosures, for example, if the message from the Issuer was truncated.
 
-It is up to the Holder how to maintain the mapping between the Disclosures and the plaintext claim values to be able to display them to the End-User when needed.
```
bc-pi, line 836: I sorta feel like something like this sentence should remain.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -833,16 +835,14 @@ If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Othe
 
 Note that these processing steps do not yield any guarantees to the Holder about having received a complete set of Disclosures. That is, for some digest values in the Issuer-signed JWT (which are not decoy digests) there may be no corresponding Disclosures, for example, if the message from the Issuer was truncated.
```
bc-pi, line 837:
```suggestion
It is up to the Holder how to maintain the mapping between the Disclosures and the plaintext claim values to be able to display them to the user when needed.
```

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: See https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/481/files#r1817324384 'cause github won't let me make a suggestion on deleted lines.

bc-pi:
should prob add something in history just for posterity - even just a -14 with a "stuff from WGLC part deux"

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: or perhaps
```suggestion
It is up to the Holder's software how to maintain the mapping between the Disclosures and the plaintext claim values to be able to display them to the user when needed.
```

danielfett: Everything here is about the user, so I would go for the first version. I initially removed the sentence because it didn't add much value and deepens the potential Holder/user confusion. But I'm fine with keeping it or a suitable replacement.

bc-pi: I kinda presumed as much. I do think it adds some value so would like something suitable kept or replaced.

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: agree with Brian..

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: +1 to the first proposal of the addition

Sakurann (approved):
assuming two suggestions Brian and I agree on get incorporated

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann:
```suggestion
: An entity that received SD-JWTs from the Issuer and has control over them. In the context of this document, the term may refer to the actual user, the supporting hardware and software in their possession, or both.
```

bc-pi (commented):
lgtm

bc-pi (approved):

PR #525, opened 2024-11-11
Note that the Hash Function Claim value is case-sensitive

bc-pi:


danielfett (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -339,6 +339,7 @@ appear at the top level of the SD-JWT payload. It
 MUST NOT be used in any object nested within the payload. If the  `_sd_alg`
 claim is not present at the top level, a default value of `sha-256` MUST be used.
 
+The claim value is a case-sensitive string with the hash algorithm identifier.
```
Sakurann, line 342:
```suggestion
This claim value is a case-sensitive string with the hash algorithm identifier.
```

Sakurann (approved):
assuming my small nit gets accepted

bc-pi:
> assuming my small nit gets accepted

No good not goes unaccepted

PR #535, opened 2025-01-09
add Watson Ladd's suggested text with minor adaptations

bc-pi:
https://mailarchive.ietf.org/arch/msg/oauth/zC8fU5nOVHsY4S5erBlBEy2fDTI/ and https://mailarchive.ietf.org/arch/msg/oauth/ugVBj2O0hw-nuWNVTFpt0JH3yVY/ are *some* threads on the mailing list

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1672,6 +1672,8 @@ the media type is encoded as an SD-JWT.
   </author>
   <date year="2021"/>
 </front>
+</reference>
+
 <reference anchor="CL01" target="https://eprint.iacr.org/2001/019.pdf">
   <front>
     <title>An Efficient System for Non-Transferable Anonymous Credentials with Optional Anonymity Revocation</title>
@@ -1684,7 +1686,6 @@ the media type is encoded as an SD-JWT.
     <date year="2001" />
   </front>
   <seriesInfo name="Proceedings of the International Conference on the Theory and Application of Cryptographic Techniques (EUROCRYPT)" value="2001" />
-</reference>
 </reference>
 
 <reference anchor="NIST.SP.800-57pt1r5" target="https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf">
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1666,7 +1666,7 @@ the media type is encoded as an SD-JWT.
 
 <reference anchor="ISO.18013-5" target="https://www.iso.org/standard/69084.html">
 <front>
-  <title>ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving license — Part 5: Mobile driving license (mDL)  application</title>
+  <title>ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving license — Part 5: Mobile driving license (mDL) application</title>
   <author>
     <organization> ISO/IEC JTC 1/SC 17 Cards and security devices for personal identification</organization>
   </author>
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1278,9 +1278,9 @@ particular Verifiers and Issuers can know when they have seen the same
 credential no matter what fields have been disclosed, even none of them.
 This behavior may not accord with what users intuitively expect or are
 lead to expect from user interface interactions and lead to them make choices they
-would not otherwise make. Workarounds such as issuing multiple
-credentials at once and using them only one time can help for keeping
-Verifiers from linking different presentations, but cannot work for Issuers.
+would not otherwise make. Workarounds such as batch issuance as
+described above help for keeping
+Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
 This issue applies to all selective disclosure based approaches,
 including mDL and mDoc [@?ISO.18013-5], and SD-CWT [@?I-D.ietf-spice-sd-cwt].
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1255,6 +1255,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
 hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
+SD-JWT only conceals the value of claims that aren't revealed.
+It does not meet standard security notations for anonymous credentials. In
+particular Verifiers and Issuers can know when they have seen the same
+credential no matter what fields have been disclosed, even none of them.
+This behavior may not accord with what users intuitively expect or are
+lead to expect from user interface interactions and lead to them make choices they
+would not otherwise make. Workarounds such as issuing multiple
+credentials at once and using them only one time can help for keeping
+Verifiers from linking different presentations, but cannot work for Issuers.
+This issue applies to all selective disclosure based approaches,
+including mDL and mDoc [@?ISO.18013-5], and SD-CWT [@?I-D.ietf-spice-sd-cwt].
+
 ## Storage of User Data
 
 Wherever user data is stored, it represents a potential
@@ -1375,7 +1387,8 @@ Simon Schulz,
 Tobias Looker,
 Takahiko Kawasaki,
 Torsten Lodderstedt,
-Vittorio Bertocci, and
+Vittorio Bertocci,
+Watson Ladd, and
 Yaron Sheffer
 for their contributions (some of which substantial) to this draft and to the initial set of implementations.
 
@@ -1631,7 +1644,17 @@ the media type is encoded as an SD-JWT.
  <front>
   <title>JSON Web Token Claims</title>
   <author><organization>IANA</organization></author>
- </front>
+</front>
+</reference>
+
+<reference anchor="ISO.18013-5" target="https://www.iso.org/standard/69084.html">
+<front>
+  <title>ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving license — Part 5: Mobile driving license (mDL)  application</title>
+  <author>
+    <organization> ISO/IEC JTC 1/SC 17 Cards and security devices for personal identification</organization>
+  </author>
+  <date year="2021"/>
+</front>
 </reference>
 
 {backmatter}
@@ -1893,6 +1916,10 @@ data. The original JSON data is then used by the application. See
 # Document History
 
    [[ To be removed from the final specification ]]
+
+   -15
+
+   * Additions and adjustments to privacy considerations
 
    -14
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1281,8 +1281,8 @@ lead to expect from user interface interactions and lead to them make choices th
 would not otherwise make. Workarounds such as batch issuance as
 described above help for keeping
 Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
-This issue applies to all selective disclosure based approaches,
-including mDL and mDoc [@?ISO.18013-5], and SD-CWT [@?I-D.ietf-spice-sd-cwt].
+This issue applies to all salted hash-based approaches,
+including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].
 
 ## Storage of User Data
 
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1275,11 +1275,11 @@ beginning of the day).
 SD-JWT only conceals the value of claims that are not revealed.
 It does not meet security notations for anonymous credentials [@?CL01]. In
 particular, colluding Verifiers and Issuers can know when they have seen the same
-credential no matter what fields have been disclosed, even when none are disclosed.
+credential no matter what fields have been disclosed, even when none have been disclosed.
 This behavior may not align with what users naturally anticipate or are guided to
 expect from user interface interactions, potentially causing them to make decisions
-they might not otherwise make. Workarounds such as batch issuance as
-described above help with keeping
+they might not otherwise make. Workarounds such as batch issuance, as
+described above, help with keeping
 Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
 This issue applies to all salted hash-based approaches,
 including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1273,13 +1273,13 @@ hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
 SD-JWT only conceals the value of claims that are not revealed.
-It does not meet standard security notations for anonymous credentials. In
-particular Verifiers and Issuers can know when they have seen the same
-credential no matter what fields have been disclosed, even none of them.
-This behavior may not accord with what users intuitively expect or are
-lead to expect from user interface interactions and lead to them make choices they
-would not otherwise make. Workarounds such as batch issuance as
-described above help for keeping
+It does not meet security notations for anonymous credentials [@?CL01]. In
+particular, colluding Verifiers and Issuers can know when they have seen the same
+credential no matter what fields have been disclosed, even when none are disclosed.
+This behavior may not align with what users naturally anticipate or are guided to
+expect from user interface interactions, potentially causing them to make decisions
+they might not otherwise make. Workarounds such as batch issuance as
+described above help with keeping
 Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
 This issue applies to all salted hash-based approaches,
 including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].
@@ -1672,6 +1672,19 @@ the media type is encoded as an SD-JWT.
   </author>
   <date year="2021"/>
 </front>
+<reference anchor="CL01" target="https://eprint.iacr.org/2001/019.pdf">
+  <front>
+    <title>An Efficient System for Non-Transferable Anonymous Credentials with Optional Anonymity Revocation</title>
+    <author initials="J." surname="Camenisch" fullname="Jan Camenisch">
+      <organization>IBM Research</organization>
+    </author>
+    <author initials="A." surname="Lysyanskaya" fullname="Anna Lysyanskaya">
+      <organization>MIT</organization>
+    </author>
+    <date year="2001" />
+  </front>
+  <seriesInfo name="Proceedings of the International Conference on the Theory and Application of Cryptographic Techniques (EUROCRYPT)" value="2001" />
+</reference>
 </reference>
 
 <reference anchor="NIST.SP.800-57pt1r5" target="https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf">
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1272,7 +1272,7 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
 hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
-SD-JWT only conceals the value of claims that aren't revealed.
+SD-JWT only conceals the value of claims that are not revealed.
 It does not meet standard security notations for anonymous credentials. In
 particular Verifiers and Issuers can know when they have seen the same
 credential no matter what fields have been disclosed, even none of them.


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
 hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
+SD-JWT only conceals the value of claims that aren't revealed.
```
danielfett, line 1275:
```suggestion
SD-JWT only conceals the value of claims that are not revealed.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
...
+It does not meet standard security notations for anonymous credentials. In
+particular Verifiers and Issuers can know when they have seen the same
+credential no matter what fields have been disclosed, even none of them.
+This behavior may not accord with what users intuitively expect or are
+lead to expect from user interface interactions and lead to them make choices they
+would not otherwise make. Workarounds such as issuing multiple
+credentials at once and using them only one time can help for keeping
+Verifiers from linking different presentations, but cannot work for Issuers.
```
danielfett, line 1283:
```suggestion
would not otherwise make. Workarounds such as batch issuance as
described above help for keeping
Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
...
+This issue applies to all selective disclosure based approaches,
+including mDL and mDoc [@?ISO.18013-5], and SD-CWT [@?I-D.ietf-spice-sd-cwt].
```
danielfett, line 1285:
```suggestion
This issue applies to all salted hash-based approaches,
including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].
```

danielfett (approved):

Denisthemalice:
I have objections/concerns with this text.

The text states:

> "SD-JWT conceals only the values that aren't revealed. It does not
> meet standard security notations for anonymous credentials. In
> particular Verifiers and Issuers can know when they have seen the same
> credential no matter what fields have been opened, even none of them.
> This behavior may not accord with what users naively expect or are
> lead to expect from UX interactions and lead to them make choices they
> would not otherwise make. Workarounds such as issuing multiple
> credentials at once and using them only one time can help for keeping
> Verifiers from linking different showing.
> 

The text is confusing because it addresses two different topics:
Verifier-Verifier unlinkability and Issuer-Verifier Unlinkability. 

The last sentence means that problem can be addressed for the Verifier-Verifier unlinkability.
It is not necessary to use strong sentences like:

> " It does not meet standard security notations for anonymous credentials"

 which are not explained.

The sentence terminates with:

> but cannot work for Issuers.

This is the case for the Issuer-Verifier unlinkability.

At the moment there is text in section 10.1 that states :

> Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be achieved in salted-hash based selective disclosure approaches, such as SD-JWT, as the issued credential with the Issuer's signature is directly presented to the Verifier, who can forward it to the Issuer.

So the proposed text is redondant.

However, the use of the wording "colluding or compromised" does not seem to be appropriate.

The Issuer will not directly "collude" with a Verifier. However, a national authority might oblige a Verifier to reveal the SD-JWT that has been used for a given access and afterwards might also oblige the Issuer to reveal the identity of the End-User. 

This situation can only happen if both the Issuer and the Verifier memorize the SD-JWTs they have respectively issued or received.

So saying that "Issuer/Verifier unlinkability with a colluding or compromised Verifier cannot be achieved in salted-hash based selective disclosure approaches" is not appropriate.

Change into:

> Issuer/Verifier unlinkability can be achieved in salted-hash based selective disclosure approaches, such as SD-JWT, if the Verifier does not memorize the SD-JWTs that it has received.

> On the contrary, if the Verifier memorizes the SD-JWTs that it has received and if the Issuer memorizes the SD-JWTs that it has issued, then some authority might oblige a Verifier to reveal the SD-JWT that has been used for a given access and then oblige the Issuer of these SD-JWTs to reveal the identity of the End-Users.

The notions of "honest" or "dishonest" Verifier are not relevant. If a Verifier obeys to a local law, it should not be considered to be "dishonest".  By definition, an Issuer is trusted to be honest.

In section 10.1, the following text should be removed:

> In considering Issuer/Verifier unlinkability, it is important to note the potential for an asymmetric power dynamic between Issuers and Verifiers. This dynamic can compel an otherwise honest Verifier into collusion. For example, a governmental Issuer might have the authority to mandate that a Verifier report back information about the credentials presented to it. 
> Legal requirements could further enforce this, explicitly undermining Issuer/Verifier unlinkability. Similarly, a large service provider issuing credentials might implicitly pressure Verifiers into collusion by incentivizing participation in their larger ecosystem. 
> Deployers of SD-JWT must be aware of these potential power dynamics, mitigate them as much as possible, and/or make the risks transparent to the user.
> Contrary to that, Issuer/Verifier unlinkability with an honest Verifier can generally be achieved. 

If the Verifier does not memorize the SD-JWTs that it has received, then the risk can be avoided.

But there are also implications with section 10.2 , where the text states:

> All involved parties SHOULD store SD-JWTs containing privacy-sensitive data only for as long as needed, including in log files.
> 
> After Issuance, Issuers SHOULD NOT store the Issuer-signed JWT or the respective Disclosures if they contain privacy-sensitive data.

The term " privacy-sensitive data" is not defined.  A SD-JWT necessarily contains 'privacy-sensitive data'. 

The first sentence would prevent to support the Issuer/Verifier unlinkability that can be achieved. 
This first sentence should be removed.

The second sentence should be shortened and changed into:

> After Issuance, Issuers SHOULD NOT store the Issuer-signed JWT.

In section 10.2, the text continues with:

> Holders SHOULD store SD-JWTs only in encrypted form, and, wherever possible, use hardware-backed encryption in particular for the private Key Binding key. Decentralized storage of data, e.g., on user devices, SHOULD be preferred for user credentials over centralized storage. 
>
> Expired SD-JWTs SHOULD be deleted as soon as possible.
> 
> After Verification, Verifiers SHOULD NOT store the Issuer-signed JWT or the respective Disclosures if they contain privacy-sensitive data. It may be sufficient to store the result of the verification and any user data that is needed for the application.

The first recommendation is an implementation option. It should be removed.

The second recommendation might be misunderstood, as it would mean that Holders SHOULD NOT delete NON-EXPIRED SD-JWTs. It should be removed.

The third recommendation can also be misunderstood, as it would mean that Verifiers SHOULD store the Issuer-signed JWT or the respective Disclosures *if* they DO NOT contain privacy-sensitive data. 

It should be changed into:

> After Verification, Verifiers SHOULD NOT store the Issuer-signed JWT. It may be sufficient to store the result of the verification and any user data that is needed for the application.

This closes the loop: if neither the Issuer nor the Verifier stores the SD-JWT, then Issuer/Verifier unlinkability can be achieved.

wbl:
Denis, the goal of my contribution was to provide a simple summary paragraph of the entire set of considerations. In that sense it is redundant. We can add a citation to CL01, or some other paper that uses the standard security notion to explain it. I'm looking for a good one.

I think a bunch of your comment is directed at the existing text rather than this PR. But I do want to say that we can't assume erasure happens. There's a whole bunch of places where data ends up in modern architectures that aren't really visible: HTTP logs, temporary files etc. We also don't assume that people will behave in security considerations: that's the joke in 3514, so guidance that only works if the verifier or issuer behave doesn't really do anything.

bc-pi:
I am not familiar with CL01 - is it this https://eprint.iacr.org/2001/019 that you mentioned in https://mailarchive.ietf.org/arch/msg/oauth/4M0qAdKhlMvACIWPpCaBTqiAOrs/ ? 

We can add a reference to that. Or a better one, if you can find one https://mailarchive.ietf.org/arch/msg/oauth/j5FoZeisLiw0kIgiNlcYZAyIsMo/

wbl:
I think referencing that eprint and paper will do. Not sure there is a much better one out there.

wbl:
Denis, I think that comment needs to be on the list (not sure how the WG works, but that's certainly where we've had most of the important discussions)

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
...
+particular Verifiers and Issuers can know when they have seen the same
+credential no matter what fields have been disclosed, even none of them.
+This behavior may not accord with what users intuitively expect or are
+lead to expect from user interface interactions and lead to them make choices they
+would not otherwise make. Workarounds such as batch issuance as
+described above help for keeping
+Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
+This issue applies to all salted hash-based approaches,
+including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].
+
```
danielfett, line 1286:
```suggestion
SD-JWT only conceals the value of claims that are not revealed.
It does not provide Issuer/Verifier unlinkability without constraints
as schemes using zero-knowledge protocols do (for example, [@?CL01]). In
particular Verifiers and Issuers can know when they have seen the same
credential no matter what fields have been disclosed, even none of them.
This behavior may not accord with what users intuitively expect or are
lead to expect from user interface interactions and lead to them make choices they
would not otherwise make. Workarounds such as batch issuance as
described above help for keeping
Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
This issue applies to all salted hash-based approaches,
including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].

```

danielfett:
@Denisthemalice I think some of your points should go into a separate PR. I'll try and create one...

Denisthemalice:
You are right, my text change proposals should be addressed in another issue. So please create a new issue.

I propose the following title for it: 

> Issuer/Verifier unlinkability and Storage of SD-JWTs

However, these two issues are strongly related.

The revised text that is being proposed is still misleading and the rational is mentioned in this "other" issue.

A sentence like 

> " This behavior may not accord with what users intuitively expect or are lead to expect from user interface interactions and lead to them make choices they would not otherwise make" 

is not understandable and should not be included into the document.

When using one-time use digital credentials, the same benefits as systems using ZKP can be obtained when considering the whole picture, I mean, including suspension and revocation checking.

We will be able to fix this issue 535 when the "other issue" will be fixed. So please , place this issue "on hold".

danielfett:
Please see PR #540

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1648,7 +1661,17 @@ the media type is encoded as an SD-JWT.
  <front>
   <title>JSON Web Token Claims</title>
   <author><organization>IANA</organization></author>
- </front>
+</front>
+</reference>
+
+<reference anchor="ISO.18013-5" target="https://www.iso.org/standard/69084.html">
+<front>
+  <title>ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving license — Part 5: Mobile driving license (mDL)  application</title>
```
bc-pi, line 1669:
```suggestion
  <title>ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving license — Part 5: Mobile driving license (mDL) application</title>
```

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: is `[@?CL01]` just gonna work? I need to reevaluate some of my belief systems, if it does.

bc-pi: do not schemes using zero-knowledge protocols also have constraints with respect to Issuer/Verifier unlinkability? Just somewhat less constraining constraints.

Denisthemalice:
You wrote:

> Yes, if every credential has a unique serial number that's always revealed, you have a problem.


There is a technique that does not need to include either a unique serial number or a tuple of uri and index 
inside a Referenced Token. This has been explained in detail when discussing revocation checking and suspension in:

https://github.com/oauth-wg/draft-ietf-oauth-status-list/issues/209#issuecomment-2558666716

This technique is different from what :

> "People have been working on this for a long time".

wbl:
Denis, my point about serial numbers was not for you. It was in response to Brian's question about what the sign and prove techniques do. 

The link you gave is not terribly detailed. I was thinking along the lines of proving an element isn't a list, of which there are lots of protocols of varying efficiency and with varying tradeoffs. Again there is a vast literature here that everyone involved in making this spec seems to have entirely ignored.

I don't think this is a great format for having this discussion given the UI and the comparatively few people looking at it: I think you should send an email to the list.

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
 hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
+SD-JWT only conceals the value of claims that are not revealed.
+It does not meet standard security notations for anonymous credentials. In
+particular Verifiers and Issuers can know when they have seen the same
+credential no matter what fields have been disclosed, even none of them.
```
Sakurann, line 1278:
```suggestion
credential no matter what fields have been disclosed, even when none .
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
...
+This behavior may not accord with what users intuitively expect or are
+lead to expect from user interface interactions and lead to them make choices they
+would not otherwise make. Workarounds such as batch issuance as
+described above help for keeping
+Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
```
Sakurann, line 1283:
```suggestion
described above help with keeping
Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
 hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
+SD-JWT only conceals the value of claims that are not revealed.
+It does not meet standard security notations for anonymous credentials. In
```
Sakurann, line 1276: Can we have a reference to what "anonymous credentials"? do we really expect all of the readers to know what it is? this is the first time that they are mentioned in the text I believe?

draft-ietf-oauth-selective-disclosure-jwt.md
Sakurann: please combine this suggestion with the suggestions I have made above :)

wbl: @bc-pi Exactly. Saying without further conditions doesn't quite work, saying what the exact notion is is technical, pointing to a reference is the happy medium and lets us get the important part across.

draft-ietf-oauth-selective-disclosure-jwt.md
wbl: I think the idea was to add a citation to CL01, but that might not actually have the phrase in it easily findable...

danielfett: I added the citation below.

draft-ietf-oauth-selective-disclosure-jwt.md
danielfett:
```suggestion
credential no matter what fields have been disclosed, even when none are disclosed.
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
...
+particular Verifiers and Issuers can know when they have seen the same
+credential no matter what fields have been disclosed, even none of them.
+This behavior may not accord with what users intuitively expect or are
+lead to expect from user interface interactions and lead to them make choices they
+would not otherwise make. Workarounds such as batch issuance as
```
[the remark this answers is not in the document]
danielfett:
```suggestion
This behavior may not align with what users naturally anticipate or are guided to
expect from user interface interactions, potentially causing them to make decisions
they might not otherwise make. Workarounds such as batch issuance as
```

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1272,6 +1272,18 @@ time period considered appropriate (e.g., randomize `iat` within the last 24
...
+It does not meet security notations for anonymous credentials [@?CL01]. In
+particular, colluding Verifiers and Issuers can know when they have seen the same
+credential no matter what fields have been disclosed, even when none are disclosed.
+This behavior may not align with what users naturally anticipate or are guided to
+expect from user interface interactions, potentially causing them to make decisions
+they might not otherwise make. Workarounds such as batch issuance as
+described above help with keeping
+Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
+This issue applies to all salted hash-based approaches,
+including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].
```
Sakurann, line 1285:
```suggestion
SD-JWT only conceals the value of claims that are not revealed.
It does not meet security notations for anonymous credentials [@?CL01]. In
particular, colluding Verifiers and Issuers can know when they have seen the same
credential no matter what fields have been disclosed, even when none have been disclosed.
This behavior may not align with what users naturally anticipate or are guided to
expect from user interface interactions, potentially causing them to make decisions
they might not otherwise make. Workarounds such as batch issuance, as
described above, help with keeping
Verifiers from linking different presentations, but cannot work for Issuer/Verifier unlinkability.
This issue applies to all salted hash-based approaches,
including mDL/mDoc [@?ISO.18013-5] and SD-CWT [@?I-D.ietf-spice-sd-cwt].
```

Sakurann (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1648,7 +1661,30 @@ the media type is encoded as an SD-JWT.
...
+      <organization>IBM Research</organization>
+    </author>
+    <author initials="A." surname="Lysyanskaya" fullname="Anna Lysyanskaya">
+      <organization>MIT</organization>
+    </author>
+    <date year="2001" />
+  </front>
+  <seriesInfo name="Proceedings of the International Conference on the Theory and Application of Cryptographic Techniques (EUROCRYPT)" value="2001" />
+</reference>
 </reference>
```
Sakurann, line 1688:
```suggestion
<reference anchor="ISO.18013-5" target="https://www.iso.org/standard/69084.html">
<front>
  <title>ISO/IEC 18013-5:2021 Personal identification — ISO-compliant driving license — Part 5: Mobile driving license (mDL) application</title>
  <author>
    <organization> ISO/IEC JTC 1/SC 17 Cards and security devices for personal identification</organization>
  </author>
  <date year="2021"/>
</front>
</reference>

<reference anchor="CL01" target="https://eprint.iacr.org/2001/019.pdf">
  <front>
    <title>An Efficient System for Non-Transferable Anonymous Credentials with Optional Anonymity Revocation</title>
    <author initials="J." surname="Camenisch" fullname="Jan Camenisch">
      <organization>IBM Research</organization>
    </author>
    <author initials="A." surname="Lysyanskaya" fullname="Anna Lysyanskaya">
      <organization>MIT</organization>
    </author>
    <date year="2001" />
  </front>
  <seriesInfo name="Proceedings of the International Conference on the Theory and Application of Cryptographic Techniques (EUROCRYPT)" value="2001" />
</reference>
```

Denisthemalice:
The text states:

>    SD-JWT only conceals the value of claims that are not revealed.  
>    It does not meet security notations for anonymous credentials [CL01].

[CL01] does not contain the words "notation", "anonymous"or "credentials".
Hence, this sentence is not understandable. Either remove this sentence or rephrase it.

The text continues with:

>    In particular, colluding Verifiers and Issuers can know when they
>    have seen the same credential no matter what fields have been
>    disclosed, even when none have been disclosed.  This behavior may not
>    align with what users naturally anticipate or are guided to expect
>    from user interface interactions, potentially causing them to make
>    decisions they might not otherwise make. 
> 

It is not possible to understand which "behavior may not align with what users naturally anticipate or 
are guided to expect from user interface interactions".

It is not possible either to understand why this behavior will "potentially causing them to make decisions 
they might not otherwise make".

Either remove this sentence or rephrase it.

The text continues with:

>    Workarounds such as batch  issuance, as described above, help with keeping Verifiers from
>    linking different presentations, but cannot work for Issuer/Verifier  unlinkability.  

This is untrue. If the Verifier deletes the presentations immediately after it has received them, 
presentations, including one-time-use presentations, cannot be used to link presentations with the Issuer.

Furthermore, if the Issuer is using a suspension / revocation mechanism different from the Status List Token, it does not need to use  "Referenced Tokens" and will still be able to revoke or suspend all the credentials that have been issued for the benefit of a given Holder. In such a case, the Issuer does not need to "remember" SD-JWTs that it has issued. If neither the Issuer, nor  the Verifier, keeps a record of respectively issued and received SD-JWTs, Issuer/Verifier  unlinkability can be "twice" achieved.

Since a specific mechanism to handle the suspension / revocation of a SD-JWT is outside the scope of this document, it is untrue to say that Issuer/Verifier  unlinkability "cannot work".

Either remove this sentence or rephrase it.

wbl:
Dennis, I once again would ask that you bring this to the list. This is not a great format for discussion and we need more eyes on it.

wbl:
> The text states:
> 
> > SD-JWT only conceals the value of claims that are not revealed.
> > It does not meet security notations for anonymous credentials [CL01].
> 
> [CL01] does not contain the words "notation", "anonymous"or "credentials". Hence, this sentence is not understandable. Either remove this sentence or rephrase it.

Do you have a better reference for the security properties introduced there? This sentence now reads with s/notation/properties/.
> 
> The text continues with:
> 
> > In particular, colluding Verifiers and Issuers can know when they
> > have seen the same credential no matter what fields have been
> > disclosed, even when none have been disclosed.  This behavior may not
> > align with what users naturally anticipate or are guided to expect
> > from user interface interactions, potentially causing them to make
> > decisions they might not otherwise make.
> 
> It is not possible to understand which "behavior may not align with what users naturally anticipate or are guided to expect from user interface interactions".

Happy to rephrase. I don't think that sentence is beyond the bounds of comprehensibility. The point is the user doesn't expect that exposing age lets someone unambigiously link everywhere they have shown their age: they would expect that only their age be revealed.

> 
> It is not possible either to understand why this behavior will "potentially causing them to make decisions they might not otherwise make".

Because the risks are not necessarily the same as what they expect their decisions aren't the ones they would make if they understood them.

> 
> Either remove this sentence or rephrase it.

Do you have suggestions for rephrasing.

> 
> The text continues with:
> 
> > Workarounds such as batch  issuance, as described above, help with keeping Verifiers from
> > linking different presentations, but cannot work for Issuer/Verifier  unlinkability.
> 
> This is untrue. If the Verifier deletes the presentations immediately after it has received them, presentations, including one-time-use presentations, cannot be used to link presentations with the Issuer.

There's no way for anyone to enforce that on the Verifier. It's also hard to ensure that data is erased even when trying: software stacks are complex.

> 
> Furthermore, if the Issuer is using a suspension / revocation mechanism different from the Status List Token, it does not need to use "Referenced Tokens" and will still be able to revoke or suspend all the credentials that have been issued for the benefit of a given Holder. In such a case, the Issuer does not need to "remember" SD-JWTs that it has issued. If neither the Issuer, nor the Verifier, keeps a record of respectively issued and received SD-JWTs, Issuer/Verifier unlinkability can be "twice" achieved.

We can't trust the Issuer doesn't record just because they don't have to. If men were angels, there would be no locksmiths.

> 
> Since a specific mechanism to handle the suspension / revocation of a SD-JWT is outside the scope of this document, it is untrue to say that Issuer/Verifier unlinkability "cannot work".

That is not what the sentence says. It says the mechanism in the document cannot guarantee it. Which is true: deletion on the part of other parties isn't enforced by anything here.

> 
> Either remove this sentence or rephrase it.

Denisthemalice:
The text states:

> SD-JWT only conceals the value of claims that are not revealed.
> It does not meet security notations for anonymous credentials [CL01].

[CL01] does not contain the words "notation", "anonymous"or "credentials". Hence, this sentence is not understandable. Either remove this sentence or rephrase it.

You responded:

> Do you have a better reference for the security properties introduced there? This sentence now reads with s/notation/properties/.

The same properties can be obtained with SD-JWT when using one-time-use SD-JWTs. So it is not necessary to argue that SD-JWT cannot meet properties that can be obtained using anonymous credentials.

There is no need to reference a paper.

The text continues with:

> In particular, colluding Verifiers and Issuers can know when they
> have seen the same credential no matter what fields have been
> disclosed, even when none have been disclosed. This behavior may not
> align with what users naturally anticipate or are guided to expect
> from user interface interactions, potentially causing them to make
> decisions they might not otherwise make.

I wrote: 

> 
It is not possible to understand which "behavior may not align with what users naturally anticipate or are guided to expect from user interface interactions".

You wrote: 

> Happy to rephrase. I don't think that sentence is beyond the bounds of comprehensibility. 
> The point is the user doesn't expect that exposing age lets someone unambigiously link everywhere they have shown their age: they would expect that only their age be revealed.

This argument is incorrect. If one-time-use SD-JWTs are used, then verifier/ verifier unlinkability is de facto supported.
It is not possible either to understand why this behavior will "potentially causing them to make decisions they might not otherwise make".

You wrote:

> Because the risks are not necessarily the same as what they expect their decisions aren't the ones they would make if they understood them.

Before an End-User presents a SD-JWT to a Verifier, the Verifier shall indicate which claims it would like to be revealed. The End-User shall either consent or refuse to consent. In addition, the End-User may be willing to only use one-time-use SD-JWTs. If the Holder is not configured in this way, the End-User should abort the transaction. 

You wrote:

> Do you have suggestions for rephrasing.

No. I only have a suggestion for deleting this sentence.


The text continues with:

> Workarounds such as batch issuance, as described above, help with keeping Verifiers from
> linking different presentations, but cannot work for Issuer/Verifier unlinkability.

I wrote:

> This is untrue. If the Verifier deletes the presentations immediately after it has received them, presentations, including one-time-use presentations, cannot be used to link presentations with the Issuer.

You wrote:

> There's no way for anyone to enforce that on the Verifier. It's also hard to ensure that data is erased even when trying: software stacks are complex.

The key wording "cannot work" is incorrect. Issuer/Verifier unlinkability *can* be supported in specific contexts. If a Verifier indicates in its practices statements that it deletes the received SD-JWTs once they are processed and if an auditor discovers later on that it does not do it, then it can go out of its business. The same applies for an Issuer.

I wrote:

> Furthermore, if the Issuer is using a suspension / revocation mechanism different from the Status List Token, it does not need to use "Referenced Tokens" and will still be able to revoke or suspend all the credentials that have been issued for the benefit of a given Holder. In such a case, the Issuer does not need to "remember" SD-JWTs that it has issued. If neither the Issuer, nor the Verifier, keeps a record of respectively issued and received SD-JWTs, Issuer/Verifier unlinkability can be "twice" achieved.

You wrote:

> We can't trust the Issuer doesn't record just because they don't have to. If men were angels, there would be no locksmiths.

See above, the case where an Issuer would lie.

I wrote:

> Since a specific mechanism to handle the suspension / revocation of a SD-JWT is outside the scope of this document, it is untrue to say that Issuer/Verifier unlinkability "cannot work".

You wrote:

> That is not what the sentence says. It says the mechanism in the document cannot guarantee it. Which is true: deletion on the part of other parties isn't enforced by anything here.

The sentence did not say " cannot guarantee it", but said "cannot work". This is not the same.

wbl:
On Thu, Jan 23, 2025, 2:08 AM Denisthemalice ***@***.***>
wrote:

> The text states:
>
> SD-JWT only conceals the value of claims that are not revealed.
> It does not meet security notations for anonymous credentials [CL01].
>
> [CL01] does not contain the words "notation", "anonymous"or "credentials".
> Hence, this sentence is not understandable. Either remove this sentence or
> rephrase it.
>
> You responded:
>
> Do you have a better reference for the security properties introduced
> there? This sentence now reads with s/notation/properties/.
>
> The same properties can be obtained with SD-JWT when using one-time-use
> SD-JWTs. So it is not necessary to argue that SD-JWT cannot meet properties
> that can be obtained using anonymous credentials.
>

Dennis,

This is completely wrong.

Your entire argument is premised on the idea that we can be assured the
Issuer will forget. Real attackers attack: they don't obey the rules.
That's why we have cryptography.

There is no way to enforce this sort of forgetting in the document, and if
you say we should use a TEE would need a massive amount of text saying how.
It would also create massive deployability issues.

In the real world enterprise and other architectures have a vast number of
places where data can be recorded. Even if an Issue wants to be sure they
forget, it's very hard to enforce.

Denisthemalice:
Section 10.2 (Storage of User Data) states: 

> All involved parties SHOULD NOT store SD-JWTs longer than strictly needed, including in log files.

Depending upon the context of use, this recommendation should or should not be followed since there exist cases where the Issuer-Verifier unlinkability property will prevent fighting against crime or money laundering.

In these cases, the Verifier should not memorize the whole SD-JWTs that it receives, but a unique reference specific to the SD-JWT together with an identifier of the Issuer. Some authority might request to a Verifier to reveal the identifier of the Issuer and this unique reference. Then after, the authority might communicate it to the Issuer and request the Issuer to reveal the identity of the End-User corresponding to that unique reference.

While technically, the Issuer-Verifier unlinkability property sounds good from a privacy point  of view, the privacy of the individuals needs to be balanced against the interests of governments or of international organizations, like Interpol, to fight against crime or money laundering.

wbl:
And that's not really relevant to my text addition.

Issue #530, opened 2024-11-20
Missing procedures for Holder to validate disclosures received from Issuer


rohanmahy:
In Section 7.2 "Processing by the Holder" there is no mention of the validation that a Holder needs to do to validate an Issued SD-JWT. The holder needs to a) look at all the disclosures and validate that the claims are acceptable to the Holder (the Issuer isn't claiming something incorrect), and b) make sure that there is a disclosure for each claim object after recursively replacing any `_sd` object or in any object with the reserved object key with its disclosed claim.

Sakurann:
7.2 is in addition to the checks in 7.1, just like 7.3.
7.3 says
> Upon receiving a presentation from a Holder, in the form of either an SD-JWT or an SD-JWT+KB, in addition to the checks outlined in [Section 7.1](https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#sd_jwt_verification), Verifiers need to ensure that
but 7.2 can be clearer on this. will clarify this.

bc-pi:
https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#section-7.1-4.5 in 7.1 covers (I think) "b)". And "a)" would be an implementation choice/detail that could maybe be mentioned but wouldn't be a normative requirement.

rohanmahy:
> https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-14.html#section-7.1-4.5 in 7.1 covers (I think) "b)". And "a)" would be an implementation choice/detail that could maybe be mentioned but wouldn't be a normative requirement.

Regarding steps 4 and 5, I think these are just plain wrong in the Holder verification case. These steps should be moved from 7.1 to 7.3. If the Holder does not have all the disclosures, they shouldn't use the SD-JWT. It is meaningless to say the Holder *removes* array elements and `_sd` keys from the payload, (because they can't send a payload to the Verifier without these elements). The Holder also cannot check for kleptographic attacks where the  Issuer and Verifier collude.

[...]

PR #541, opened 2025-01-13
Try to address Rohan's comments

danielfett:
Fix Issue #530

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -800,7 +800,7 @@ or a Verifier needs to ensure that:
  * all Disclosures are valid and correspond to a respective digest value in the Issuer-signed JWT (directly in the payload or recursively included in the contents of other Disclosures).
 
 The Holder or the Verifier MUST perform the following (or equivalent) steps when receiving
-an SD-JWT:
+an SD-JWT to validate the SD-JWT and extract the payload:
 
 1. Separate the SD-JWT into the Issuer-signed JWT and the Disclosures (if any).
 2. Validate the Issuer-signed JWT:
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -837,12 +837,19 @@ If any step fails, the SD-JWT is not valid, and processing MUST be aborted. Othe
...
+
 ## Processing by the Holder  {#holder_verification}
 
 The Issuer MUST provide the Holder an SD-JWT, not an SD-JWT+KB.  If the Holder
 receives an SD-JWT+KB, it MUST be rejected.
 
-For presentation to a Verifier, the Holder MUST perform the following (or equivalent) steps:
+When receiving an SD-JWT, the Holder MUST do the following:
+
+ 1. Process the SD-JWT as defined in (#sd_jwt_verification) to validate the presentation and extract the payload.
```
paulbastian, line 849: If this means the holder receives it during issuance, then I think presentation is the wrong wording here

danielfett: Correct, that was copy paste from below. 
```suggestion
 1. Process the SD-JWT as defined in (#sd_jwt_verification) to validate the SD-JWT and extract the payload.
```

bc-pi: I agree with @paulbastian --- for better or worse the term "presentation" has come to have some specific connotations in the world of digital verifiable credentials that don't align with the holder receiving the thing.

bc-pi:
```suggestion
 1. Process the SD-JWT as defined in (#sd_jwt_verification) to validate it and extract the payload.
```

bc-pi: Either suggestion above would be fine but using "SD-JWT" twice in the one sentence felt a little too redondant.

bc-pi (approved):

Sakurann (approved):

bc-pi:
Two approvals. Open for a little while. Good to go.

PR #543, opened 2025-01-17
editorial cleanup

wbl:
This seems to have gotten dropped in some editorial passes

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1282,7 +1282,7 @@ hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
 SD-JWT only conceals the value of claims that are not revealed.
-It does not meet security notations for anonymous credentials [@?CL01]. In
+It does not meet the standard security notations for anonymous credentials [@?CL01]. In
 particular, colluding Verifiers and Issuers can know when they have seen the same
 credential no matter what fields have been disclosed, even when none have been disclosed.
 This behavior may not align with what users naturally anticipate or are guided to
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -1282,7 +1282,7 @@ hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
 SD-JWT only conceals the value of claims that are not revealed.
-It does not meet the standard security notations for anonymous credentials [@?CL01]. In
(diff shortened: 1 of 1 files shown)


bc-pi (changes requested):
Upon consultation and review yesterday with the editors (@Sakurann, @danielfett and myself) agreed that we preferred the text omitting "the standard".

wbl:
The current sentence reads "It does not meet security notations for anonymous credentials [@?CL01]."  Probably a typo introduced by me, that should read notions. I'll assume we fix that. A reader is left wondering why the security notions it does not meet are relevant. There needs to be some indication of why these properties matter, and importantly that it cannot be dropped in where other schemes could be used.

(Maybe properties is clearer overall: I think we should change to that on reading this again)

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1282,7 +1282,7 @@ hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
 SD-JWT only conceals the value of claims that are not revealed.
-It does not meet security notations for anonymous credentials [@?CL01]. In
+It does not meet the standard security notations for anonymous credentials [@?CL01]. In
```
bc-pi, line 1285:
```suggestion
It does not meet the security notions for anonymous credentials [@?CL01]. In
```

bc-pi, line 1285: or
```suggestion
It does not meet the security properties for anonymous credentials [@?CL01]. In
```

bc-pi:
> The current sentence reads "It does not meet security notations for anonymous credentials [@?CL01]." Probably a typo introduced by me, that should read notions. I'll assume we fix that. A reader is left wondering why the security notions it does not meet are relevant. There needs to be some indication of why these properties matter, and importantly that it cannot be dropped in where other schemes could be used.
> 
> (Maybe properties is clearer overall: I think we should change to that on reading this again)

some possibilities suggested on the text

bc-pi (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -1282,7 +1282,7 @@ hours and calculate `exp` accordingly) or rounded (e.g., rounded down to the
 beginning of the day).
 
 SD-JWT only conceals the value of claims that are not revealed.
-It does not meet security notations for anonymous credentials [@?CL01]. In
+It does not meet the security properties for anonymous credentials [@?CL01]. In
```
bc-pi, line 1285:
```suggestion
It does not meet the security properties of anonymous credentials [@?CL01]. In
```

PR #564, opened 2025-04-04
Align PID example with latest ARF

danielfett:


bc-pi:
https://drafts.oauth.net/oauth-selective-disclosure-jwt/danielfett/update-pid-example-arf-latest/draft-ietf-oauth-selective-disclosure-jwt.html#name-sd-jwt-based-verifiable-cre for those playing along at home

bc-pi (approved):

PR #573, opened 2025-05-05
try to improve some language around exactly what bytes get base64url encoded

bc-pi:
Inspired by https://mailarchive.ietf.org/arch/msg/oauth/hd1CKEIdVodkRKmdNx-vhlTrLt0/ this PR attempts to improve some language around exactly what bytes get base64url encoded

### DIFF
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -428,7 +428,7 @@ For each claim that is an array element and that is to be made selectively discl
    1. The salt value as described in (#disclosures_for_object_properties).
    2. The array element that is to be hidden. This value can be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
 
-The Disclosure string is created by base64url-encoding the UTF-8 byte sequence of the resulting JSON array's string representation as described in (#disclosures_for_object_properties). The same considerations regarding
+The Disclosure string is created by base64url-encoding the UTF-8 byte sequence of the resulting JSON array as described in (#disclosures_for_object_properties). The same considerations regarding
 variations in the result of the JSON encoding apply.
 
 For example, a Disclosure for the second element of the `nationalities` array in the following JWT Claims Set:
--- a/draft-ietf-oauth-selective-disclosure-jwt.md
+++ b/draft-ietf-oauth-selective-disclosure-jwt.md
@@ -43,14 +43,14 @@ organization="Ping Identity"
 
 .# Abstract
 
-This specification defines a mechanism for the selective disclosure of individual elements of a JSON-encoded data structure used as the payload of a JSON Web Signature (JWS).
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-selective-disclosure-jwt.md
```diff
@@ -429,7 +428,7 @@ For each claim that is an array element and that is to be made selectively discl
    1. The salt value as described in (#disclosures_for_object_properties).
    2. The array element that is to be hidden. This value can be of any type that is allowed in JSON, including numbers, strings, booleans, arrays, and objects.
 
-The Disclosure string is created by JSON-encoding this array and base64url-encoding the byte representation of the resulting string as described in (#disclosures_for_object_properties). The same considerations regarding
+The Disclosure string is created by base64url-encoding the UTF-8 byte sequence of the resulting JSON array's string representation as described in (#disclosures_for_object_properties). The same considerations regarding
```
cabo, line 431: Please don't use "string representation" unless strings are involved.
This is one place where "JSON encoding" would make sense.

bc-pi: It is JSON so I'm not sure "JSON encoding" does make sense.

bc-pi: The language is hard.

bc-pi, line 431: this better or worse? or are we miscommunicating? 
```suggestion
The Disclosure string is created by base64url-encoding the UTF-8 byte sequence of the resulting JSON array representation as described in (#disclosures_for_object_properties). The same considerations regarding
```

bc-pi: There are lots of strings in involved. So to maybe to phrase it differently, what would you call the thing here https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-18.html#section-5.1-9.3 and in many other places after the label `Contents: `?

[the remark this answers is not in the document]
OR13: To me a JSON array is a string.

OR13 (approved):

draft-ietf-oauth-selective-disclosure-jwt.md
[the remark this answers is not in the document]
bc-pi: that wording would be okay with me

bc-pi:
Open for more than a week. One approval. Some comments addressed. Telechat fast approaching.

draft-ietf-oauth-selective-disclosure-jwt.md
bc-pi: overcome by events

bc-pi: overcome by events

PR #576, opened 2025-05-22
changes aimed at addressing one IESG member's feedback

bc-pi:
and that IESG member is @OR13


OR13 (approved):

OR13:
This covers one part of the discuss, but I still want to chat about the scenario described here:

https://mailarchive.ietf.org/arch/msg/oauth/yVX3ku5P4O-3gLinWfAsj86Lj8g/

> A relying party MUST reject SD-JWT+KB when SD-JWT or KB-JWT contain audience values and none of the audience  values are associated with the relying party.

Higher bandwidth should resolve this, and its possible this is obvious to folks already, and so redundant to the current requirements.

bc-pi:
Yep, I noted our coming higher bandwidth discussion https://mailarchive.ietf.org/arch/msg/oauth/kfFRfttW9kLhvQ0lyR6oMnuxrMQ/

OR13 (approved):

OR13 (approved):

selfissued (approved):

PR #580, opened 2025-05-29
fix text that was out of sync with the associated example

bc-pi:


bc-pi (commented):
"Pull request authors can’t approve their own pull request."

PR #583, opened 2026-01-10
Add Haskell sd-jwt link to README

yaronf:


bc-pi (approved):
Oh wow, I did one very small project with Haskell in college. Haven't seen much of it sense. 

I did happen to notice the note "Note: RS256 (RSA-PKCS#1 v1.5) is deprecated per draft-ietf-jose-deprecate-none-rsa15 due to padding oracle attack vulnerabilities" in there https://hackage.haskell.org/package/sd-jwt-0.1.0.0#readme which is contrary to the content of the draft as noted it the draft itself https://datatracker.ietf.org/doc/html/draft-ietf-jose-deprecate-none-rsa15-03#section-1-3,  

> Note that RSA signatures using PKCS#1 version 1.5 padding (RS256,
>   RS384, and RS512) are unchanged by this specification and can still
>  be used.

yaronf:
Haskell is interesting, but honestly I wouldn't have been able to do it without the AI.

Thanks for pointing out my mistake on PKCS1, fixed in a patch release.

